PR #50, opened 2025-12-17
fix(capture): Use display-based capture for window screenshots

### LINKED ISSUE #49: iOS Simulator window capture returns black image
## Problem

When using Peekaboo to capture iOS Simulator windows via `--app Simulator` or `--pid`, the captured image is completely black.

## Root Cause

The iOS Simulator renders its content through Metal/GPU compositing. When `SCContentFilter(desktopIndependentWindow:)` is used, it attempts to capture the window's backing store directly, but Simulator's GPU-rendered content bypasses this backing store, resulting in a black image.

## Reproduction

```bash
# Start iOS Simulator with any device
xcrun simctl boot 'iPhone 15 Pro'

# Attempt capture - returns black image
peekaboo capture live --app Simulator --duration 5 --path /tmp/sim_frames
```

## Verified Solution

Using `SCContentFilter(display:including:[window])` with `sourceRect` set to the window frame successfully captures the Simulator content:

```swift
// This works:
let filter = SCContentFilter(display: display, including: [window])
config.sourceRect = window.frame

// This returns black for Simulator:
let filter = SCContentFilter(desktopIndependentWindow: window)
```

## Proposed Fix

Detect when a captured image is essentially black and fall back to the display-based capture method. This maintains compatibility with normal windows while fixing GPU-rendered windows like Simulator.

## Environment

- macOS 15.2
- Peekaboo 3.0.0
- iOS Simulator (any version)

bheemreddy-samsara:
## Summary

Fixes #49 - iOS Simulator window capture returns black image.

## Problem

When capturing windows using `SCContentFilter(desktopIndependentWindow:)`, GPU-rendered windows like iOS Simulator return black images because they render through Metal/GPU compositing that bypasses the standard window backing store.

## Solution

Switch to `SCContentFilter(display:including:[window])` for all window captures. This approach:

- Works reliably for **all** windows including GPU-rendered ones
- The `including:` filter ensures only the target window is captured, even if occluded
- Simpler implementation than the previous approach (no fallback logic required)

## Changes

```swift
// Before (fails for Simulator):
let filter = SCContentFilter(desktopIndependentWindow: window)

// After (works for all windows):
let filter = SCContentFilter(display: display, including: [window])
config.sourceRect = window.frame
```

## Testing

Verified iOS Simulator windows now capture correctly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 ### Fixed
 - Visualizer previews now respect their full duration before fading out; overlays no longer disappear in ~0.3s regardless of requested timing.
 - App resolution now prioritizes exact name matches over bundleID-contains matches, preventing `--app Safari` from accidentally matching helper processes with “Safari” in their bundle ID.
+- ScreenCaptureKit window capture no longer returns black frames for GPU-rendered windows (notably iOS Simulator), and display-bound crops now use display-local `sourceRect` coordinates on secondary monitors (thanks @bheemreddy-samsara).
 - `peekaboo see` is now bounded for “single action” use:
   - Without `--analyze`, the overall wall-clock timeout is **10 seconds**.
   - `--json` output skips menubar enumeration unless `--verbose` is set, and timeouts surface as `TIMEOUT` exit codes instead of silent hangs.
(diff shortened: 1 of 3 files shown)


bheemreddy-samsara:
@steipete Can i get a review

Issue #77, opened 2026-02-20
peekaboo no permission in openclaw.


taozhiyuai:
<img width="487" height="310" alt="Image" src="https://github.com/user-attachments/assets/9806d6b1-186b-473d-9cb3-d1e444cbc6ca" />
<img width="424" height="145" alt="Image" src="https://github.com/user-attachments/assets/fa72dd10-503e-46c5-ab1f-e610f4d9904b" />
<img width="664" height="427" alt="Image" src="https://github.com/user-attachments/assets/1c2f4014-0827-4e74-aab3-1813f1c75259" />

<img width="1148" height="1137" alt="Image" src="https://github.com/user-attachments/assets/df8455f7-d4c2-44da-a452-d78dee30383d" />

already grant in MAC system setting. but reported no permission

AkashKumar7902:
@taozhiyuai how did you grant permissions ?

steipete:
Shipped a CLI/docs fix in e1f80be4.

What changed:
- `peekaboo permissions status` now prints a Bridge-specific hint when Screen Recording is missing on the selected host app.
- `peekaboo bridge status` now surfaces the same hint when a probed host reports `SR=N`.
- Docs now call out the OpenClaw/subprocess workaround explicitly:

```bash
peekaboo see --mode screen --screen-index 0 --no-remote --capture-engine cg --json
```

Root cause is that TCC applies to the process doing the capture; when Bridge is selected, the host app needs Screen Recording even if the parent OpenClaw/Node process already has it.

Issue #75, opened 2026-02-18
Screen Recording permission check fails on macOS 26 when spawned by Node.js


Orionation:
## Environment
- **macOS version:** 26.3 (Build 25D125)
- **peekaboo version:** 3.0.0-beta3 (main/69376fa4-dirty)
- **Installation:** Homebrew (steipete/tap/peekaboo)

## Problem
`peekaboo permissions` reports "Screen Recording: Not Granted" when run as a child process of Node.js, even when all relevant checkboxes are enabled in System Settings.

### What's enabled in System Settings → Screen & System Audio Recording:
- ✅ node
- ✅ OpenClaw (the Node.js app)
- ✅ peekaboo
- ✅ Terminal

### Actual behavior:
```
$ peekaboo permissions
Source: local runtime
Screen Recording (Required): Not Granted
Accessibility (Required): Granted
```

Running `peekaboo see` fails with permission error.

## Workaround
Running peekaboo via Terminal.app works perfectly:
```bash
osascript -e 'tell application "Terminal" to do script "peekaboo see --path /tmp/test.png"'
```

This confirms:
1. Terminal.app has proper Screen Recording permission
2. peekaboo inherits it correctly from Terminal
3. The issue is specific to Node.js as parent process

## Hypothesis
macOS 26 may have changed how Screen Recording permissions propagate to child processes. The permission check in peekaboo may need updating for macOS 26 compatibility.

## Additional Notes
- Accessibility permission works correctly (Granted, and AppleScript automation works)
- Native `/usr/sbin/screencapture` also fails when run through Node.js
- Rebooting doesn't help
- Toggling permissions off/on doesn't help
- `tccutil reset ScreenCapture` and re-granting doesn't help

## Workaround Script
For anyone hitting this, wrap peekaboo calls through Terminal.app:
```bash
#!/bin/bash
osascript -e "tell application \"Terminal\" to do script \"peekaboo see --path '\$1'; exit\""
```

trietphan:
## Reproduction from OpenClaw node runner (macOS 15.4, M4 MacBook Air)

Hitting this exact issue when OpenClaw spawns Peekaboo commands via its node runner.

### Environment
- macOS 15.4 (Sequoia), Apple M4
- Peekaboo 3.0.0-beta3 (Homebrew)
- OpenClaw gateway spawning commands via paired macOS node
- Screen Recording + Accessibility permissions: **granted** (confirmed via `peekaboo permissions --json`)

### What fails
[...]

Issue #52, opened 2025-12-27
Swift continuation leak causes see/image/permissions commands to hang on macOS Tahoe


dillycone:
## Environment
- **Peekaboo version:** 3.0.0-beta2 (installed via Homebrew, built from source 2025-12-27)
- **macOS version:** 26.2 (Tahoe beta, Build 25C5037j)
- **Bridge status:** local (in-process)

## Issue
Commands that require screen recording permissions (`see`, `image`, `permissions`) hang indefinitely with a Swift async continuation leak error:

```
SWIFT TASK CONTINUATION MISUSE: _createCheckedThrowingContinuation(_:) leaked its continuation without resuming it. This may cause tasks waiting on it to remain suspended forever.
```

## Commands Affected
- `peekaboo see --app Safari` ❌ hangs
- `peekaboo image --app Safari --format png --path /tmp/test.png` ❌ hangs  
- `peekaboo permissions` ❌ hangs

## Commands Working
- `peekaboo list` ✅
- `peekaboo app launch/quit` ✅
- `peekaboo hotkey` ✅
- `peekaboo type` ✅
- `peekaboo bridge status` ✅

## Steps to Reproduce
```bash
# This hangs with continuation leak error
peekaboo see --app Safari

# Even with --no-remote flag
peekaboo see --app Safari --no-remote
```

## Analysis
The continuation leak suggests an async code path in the screen recording permission check or CGWindowListCreateImage wrapper where `continuation.resume()` is never called. This might be Tahoe-specific behavior where the permission APIs behave differently.

## Related
- Issue #51 (Tahoe permissions issue) - related but different symptom
- Issue #42 (see gets stuck) - was closed as fixed, but this is a different manifestation

## Workaround
AppleScript can be used for basic automation:
```bash
osascript -e 'tell application "Safari" to activate'
```

Jeehut:
Confirming on Peekaboo 3.0.0-beta4 (main/8659b70d, built 2026-04-15) running on macOS 26.4.1 (Tahoe):

- **The continuation leak is fixed** for `image` and `permissions` — both complete in a few hundred ms. 🎉
- **`see` still hangs** on SwiftUI apps, but the symptom has changed. With the on-demand daemon started (`peekaboo daemon start`), the capture phase completes cleanly in ~200ms and the PNG is saved, but the subsequent `element_detection` phase hangs and hard-aborts at exactly 25003ms regardless of `--timeout-seconds 45/60`. Output includes:

```
[...]

PR #78, opened 2026-02-25
docs: add agent bridge troubleshooting guide for OpenClaw/Node.js contexts

trietphan:
## Summary

Adds a troubleshooting guide for AI agent frameworks (OpenClaw, Claude Code, etc.) that spawn Peekaboo via Node.js runtimes and hit bridge capture failures.

## Problem

When Peekaboo is spawned by an agent framework's Node.js process, the bridge daemon's `captureWindow`/`captureFrontmost` operations fail with `PeekabooBridgeErrorEnvelope` or `Failed to start stream due to audio/video capture failure`.

This happens because ScreenCaptureKit TCC grants don't propagate from the parent terminal/agent process to the bridge daemon process on macOS 15+.

Non-capture commands (`click`, `type`, `hotkey`, `list`) work fine through the bridge.

## Workaround documented

```bash
peekaboo daemon stop
peekaboo see --mode screen --screen-index 0 --capture-engine cg --no-remote --json
```

CoreGraphics engine bypasses SCKit and works reliably in agent-spawned contexts.

## Related issues

- #75 — Screen Recording permission check fails on macOS 26 when spawned by Node.js
- #77 — Peekaboo no permission in OpenClaw
- #52 — Swift continuation leak causes see/image/permissions commands to hang

## Testing

Tested on:
- macOS 15.4, Apple M4 MacBook Air
- Peekaboo 3.0.0-beta3 (Homebrew)
- OpenClaw gateway with paired macOS node

### DIFF
--- a/docs/agent-bridge-troubleshooting.md
+++ b/docs/agent-bridge-troubleshooting.md
@@ -1,0 +1,71 @@
+# Agent Bridge Troubleshooting
+
+When running Peekaboo from AI agent frameworks (OpenClaw, Claude Code, etc.) that spawn commands via Node.js or similar runtimes, you may encounter capture failures even when permissions are granted.
+
+## Common Issue: Bridge Capture Failure
+
+### Symptoms
+
+```
+INTERNAL_SWIFT_ERROR: Failed to start stream due to audio/video capture failure
+PeekabooBridgeErrorEnvelope error 1
+```
+
+These errors occur on commands that use window/screen capture (`see`, `image --mode window`) when routed through the Peekaboo bridge daemon.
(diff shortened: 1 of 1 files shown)


steipete:
Thanks for the PR, @trietphan. I’m going to close this one after a deeper review.

The write-up is helpful, but it is tied pretty tightly to a specific  workaround path and Node/OpenClaw context, so it’s likely to age badly and mislead current users if merged as-is.

I’d rather fold this into a more general, current troubleshooting guide later instead of landing a version-bound document. Thanks for putting the investigation together.

PR #86, opened 2026-03-08
fix(mac): avoid persisting settings while loading

### LINKED ISSUE #80: add support for local models
Using the macOS app there are options to configure local models but they are just not honored.

```
peekaboo agent
Error: Agent service not available. Please set OPENAI_API_KEY, ANTHROPIC_API_KEY, or GEMINI_API_KEY.
```

Furthermore, "Enable agent mode" doest stay enabled on  the GUI.

Image

After enabling a custom provider multiple "Default" models are selected:

Image

Even though a custom provider is selected:

Image

In a new Session window OpenAI or Anthropic is ALWAYS requested:

Image
When I set fake keys a "test" for both models and have the "Default" model from the custom provider selected it still wont work

Image

and OpenAI is used by default

Image

selecting LLaVa for Ollama does the same thing 

Image

0xble:
Fixes #80.

Prevents the macOS settings object from saving UserDefaults, credentials, or config.json while it is still hydrating from config-backed provider state. That avoids the launch-time partial write that was resetting unrelated UI settings and rewriting an existing config file.

I also added a regression test for config-backed init preserving both the existing config and unrelated UI defaults. Built `Apps/Mac` locally. The current Mac package test target still has unrelated upstream compile failures in `EndToEndTests` and `StatusBarControllerTests`, so the new test is included in the package but that target does not fully run clean on `main` today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Fixed
+- macOS settings hydration no longer persists config-backed values while loading; thanks @0xble for [#86](https://github.com/steipete/Peekaboo/pull/86).
 - CLI agent runtime now prefers local execution by default; thanks @0xble for [#83](https://github.com/steipete/Peekaboo/pull/83).
 - Remote `peekaboo see` element detection now uses the command timeout instead of the bridge client's shorter socket default; thanks @0xble for [#89](https://github.com/steipete/Peekaboo/pull/89).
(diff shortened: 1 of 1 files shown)


steipete:
Landed via temp rebase onto main.

- Gate: `pnpm run lint && pnpm run build:cli && pnpm run test:safe`
- Land commit: 3e13e906fb59fefc4d196ea0a48ebf2943407faf
- Merge commit: af55bad8801c3121574775824da5810ae8ce8ddf

Thanks @0xble!

PR #96, opened 2026-03-20
feat(cli): add completions

jkker:
## Summary

Add `peekaboo completions [shell]` so the CLI can emit zsh, bash, and fish completion scripts generated from Commander metadata. Commands, aliases, and value suggestions stay in sync with the shipped CLI — no handwritten completion tables.

### Source (`Apps/CLI/Sources/PeekabooCLI/CLI/Completions/`)
- **CompletionModel.swift** (279 LOC) — shared model types, value catalog, dispatcher, protocol
- **BashCompletionRenderer.swift** (221 LOC) — self-contained bash completion script renderer
- **ZshCompletionRenderer.swift** (193 LOC) — zsh compdef + dynamic helpers renderer
- **FishCompletionRenderer.swift** (187 LOC) — fish complete callback renderer

### Command
- **CompletionsCommand.swift** — `peekaboo completions [shell]` with auto-detection from \$SHELL
- **CompletionsCommand+CommanderMetadata.swift** — registration and binder

### Tests (`Apps/CLI/Tests/CoreCLITests/CompletionsCommandTests.swift`)
- Shell resolution (explicit names, paths, variants like `bash-old`)
- Metadata extraction (commands, aliases, argument choices)
- Script shape validation (function names, compdef, complete -c)
- **Shell-parse smoke tests**: `bash -n`, `zsh -n`, `fish --no-execute` validate generated scripts parse without syntax errors

### Docs
- `docs/commands/completions.md` — user-facing setup and troubleshooting
- `docs/dev/completions.md` — architecture and extension guide

## Validation

```bash
# Lint (0 issues)
swiftformat --lint Apps/CLI/Sources/PeekabooCLI/CLI/Completions/ \\
  Apps/CLI/Sources/PeekabooCLI/Commands/Core/CompletionsCommand.swift \\
  Apps/CLI/Tests/CoreCLITests/CompletionsCommandTests.swift

# Tests (296 tests, 53 suites — all pass)
swift test --package-path Apps/CLI -Xswiftc -DPEEKABOO_SKIP_AUTOMATION --no-parallel

# End-to-end shell validation
peekaboo completions bash | bash -n  # ✓
peekaboo completions zsh  | zsh -n   # ✓
```

## Addresses review feedback
- Split the original 877-line `CompletionScriptSupport.swift` into 4 focused files (all under 280 LOC)
- Added shell-parse smoke tests that validate generated scripts actually parse/load in target shells
- Fixed shell resolution for variant paths like `/bin/bash-old` and `/bin/bash-`


jkker:
any thoughts @steipete @MSch @0xble ? hoping to get this merged if you'd like

steipete (changes requested):
Thanks for the detailed completions work. Blocking issue: `Apps/CLI/Sources/PeekabooCLI/CLI/Completions/CompletionScriptSupport.swift` is 877 lines, which violates the repo guidance to keep files under roughly 500 LOC and makes the shell renderers hard to review safely. Please split the shared model/catalog from the bash/zsh/fish renderers into separate files and add at least smoke validation that the generated scripts parse/load in the target shells, not only Swift string-shape tests. After that this is much easier to review on the merits.

jkker (commented):
Thanks for the feedback! I've addressed both points:

**File split** — The 877-line `CompletionScriptSupport.swift` is now split into 4 focused files:
- `CompletionModel.swift` (279 LOC) — model types, value catalog, dispatcher
- `BashCompletionRenderer.swift` (221 LOC)
- `ZshCompletionRenderer.swift` (193 LOC)
- `FishCompletionRenderer.swift` (187 LOC)

**Shell-parse smoke tests** — Added `bash -n` and `zsh -n` validation tests (plus `fish --no-execute` when fish is available). These confirm the generated scripts have no syntax errors.

Also fixed the shell variant resolution (`/bin/bash-old`, `/bin/bash-`) and ran SwiftFormat to match project style. Full test suite passes (296 tests, 53 suites).

jkker:
## RFC: Completions via swift-argument-parser (future direction)

While working on this PR, I did a deep-dive into completion generation approaches. Wanted to share findings and open a conversation about whether `swift-argument-parser` (SAP) could complement Commander for the completions use-case specifically.

### Context

The current approach (this PR) generates completion scripts from Commander descriptor metadata in Swift. It works well and keeps the SSOT principle, but it means we maintain ~880 LOC of shell-script renderers that need to track bash/zsh/fish semantics.

### What SAP offers for completions

Apple's swift-argument-parser has battle-tested completion generators (maintained since 2020, used by hundreds of CLI tools):
- Handles edge cases we'd need to discover over time (repeating options, mutually exclusive flags, custom completion callbacks, async completions)
- Supports `--generate-completion-script` as a built-in subcommand
- Actively maintained by Apple's tooling team with per-shell regression tests

### Challenges & why Commander stays

Commander is purpose-built for Peekaboo's needs — runtime binding, the descriptor/signature system, injected runtime flags, and help rendering are all tightly integrated. SAP doesn't support these patterns without significant adaptation. Migrating the CLI framework itself would be a large effort with unclear ROI.

### Possible middle ground (non-blocking, just brainstorming)

One option that doesn't require migrating away from Commander:

1. Add a build-time step that generates an SAP-compatible `CommandInfo` (the `ArgumentParserToolInfo` JSON format) from Commander descriptors
2. Feed that into SAP's completion generators as a library dependency
3. We get Apple's shell renderers for free, drop our custom renderers, and still keep Commander for everything else

This is purely additive — Commander stays as the runtime framework, SAP just becomes a code-gen dependency for the completion scripts.

### Questions for you

- Is the custom renderer approach fine for now, and we iterate if users report shell-specific bugs?
- Would you be open to exploring the SAP-as-completion-backend idea in a future PR if the maintenance burden of the renderers grows?
- Any other directions you'd prefer?

Happy to go either way — the current implementation in this PR is solid and self-contained. This is just flagging an option for the roadmap.

steipete (approved):
Follow-up review after the split/render-smoke update: no blocking correctness issues found. I also pushed a small cleanup commit to remove the new lint/compiler warnings and re-ran the completions test suite plus bash/zsh parse smoke checks.

PR #99, opened 2026-03-25
support custom ai providers in image analyze

381181295:
## what i expected

`image --analyze` should honor custom providers configured in `~/.peekaboo/config.json`, eg entries like:

- `local-proxy/gpt-5.4-mini`

that matches the documented custom-provider setup and is especially useful for local openai-compatible proxies.

## what i ran into

`image --analyze` ignored custom providers and silently fell back to built-in providers/models instead.

in practice, that meant my local proxy config never actually took effect for image analysis.

## root cause

`PeekabooAIService.parseProviderEntry` only handled built-in provider names like:

- `openai`
- `anthropic`
- `gemini`
- etc

for anything else, it returned `nil`.

there was also a cli/config mismatch: tachikoma needed to look in `.peekaboo` for the same config profile the cli uses.

## fix

- teach `PeekabooAIService` to resolve custom providers through `CustomProviderRegistry`
- map `provider/model` entries like `local-proxy/gpt-5.4-mini` into real custom provider instances
- align cli startup with the `.peekaboo` profile dir
- add a regression test covering a custom openai-compatible provider

## result

after this change, `image --analyze` can use a custom provider from config instead of falling back to built-ins.

## note

depends on the tachikoma custom-provider credential/init fix first.

### DIFF
--- a/Apps/CLI/Sources/PeekabooCLI/CLI/PeekabooEntryPoint.swift
+++ b/Apps/CLI/Sources/PeekabooCLI/CLI/PeekabooEntryPoint.swift
@@ -3,6 +3,7 @@ import CoreGraphics
 import Darwin
 import Foundation
 import PeekabooCore
+import Tachikoma
 
 /// Shared entry point used by the executable target.
 @MainActor
@@ -20,6 +21,9 @@ func executePeekabooCLI(arguments: [String]) async -> Int32 {
 
     // Initialize CoreGraphics silently to prevent CGS_REQUIRE_INIT error
     _ = CGMainDisplayID()
+
+    // Keep Tachikoma's profile dir aligned with Peekaboo's config dir.
+    TachikomaConfiguration.profileDirectoryName = ".peekaboo"
 
     // Load configuration at startup
     _ = ConfigurationManager.shared.loadConfiguration()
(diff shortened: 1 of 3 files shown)


steipete (changes requested):
Thanks, this is the right area, but I would not merge this shape. It combines Peekaboo changes with a Tachikoma submodule bump, and the added `PeekabooAIServiceTests.resolvesCustomOpenAICompatibleProvider` appears to call `analyzeImageFileDetailed` rather than testing provider resolution behind a stub, so it can require a reachable OpenAI-compatible endpoint/model during the unit suite. Please split the Tachikoma prerequisite into its own upstream PR/submodule bump first, then make the Peekaboo test hermetic by asserting parsed/resolved provider metadata without a live image-analysis request.

steipete:
Thanks for the report and PR. I landed the custom-provider resolution fix directly on main in 11cf0ac7, with the regression moved into the active PeekabooTests target so it actually runs.\n\nI did not merge this branch as-is because it still carries the stale Tachikoma submodule bump and the original live image-analysis unit test shape. Main now resolves configured entries like local-proxy/model for image analysis without falling back to built-in defaults.

PR #103, opened 2026-04-04
Improve browser window focus stabilization

ZVNC28:
## Summary
- re-resolve the AX window handle after activation/space changes
- strengthen application activation before focus verification
- relax focus verification to accept frontmost-app matches for browser-style windows

## Why
Browser windows can be flaky during automation because AX window handles may go stale after activation and some web-heavy apps do not mark the target window as main quickly enough.

## Validation
- built successfully with [0/1] Planning build
Building for debugging...
[0/1] Write swift-version--1AB21518FC5DEDBE.txt
Build complete! (1.43s)

### DIFF
--- a/Core/PeekabooAutomationKit/Sources/PeekabooAutomationKit/Utilities/FocusUtilities.swift
+++ b/Core/PeekabooAutomationKit/Sources/PeekabooAutomationKit/Utilities/FocusUtilities.swift
@@ -165,38 +165,48 @@ public final class FocusManagementService {
 
     /// Focus a window by its CGWindowID
     public func focusWindow(windowID: CGWindowID, options: FocusOptions = FocusOptions()) async throws {
-        // Attempting to focus window
-
-        // Verify window exists
+        // Verify window exists before any focus work starts.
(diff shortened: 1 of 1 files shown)


steipete:
Thanks for the focus-stabilization patch. I landed the narrower version on main in 295d5207.\n\nWhat changed:\n- re-resolve the AX window handle after app activation/Space movement in Core/PeekabooAutomationKit/Sources/PeekabooAutomationKit/Utilities/FocusUtilities.swift:178\n- keep AX main-window verification, but add a stricter fallback that confirms the requested window is the topmost renderable CG window in Core/PeekabooAutomationKit/Sources/PeekabooAutomationKit/Utilities/FocusUtilities.swift:269\n- cover Chrome/browser helper-window ordering in Core/PeekabooCore/Tests/PeekabooTests/FocusUtilitiesTests.swift\n\nI avoided the broader frontmost-app-only success check because it can pass for the wrong window when a browser has multiple windows. Local proof: focused test, CLI build, Safari/Chrome menu/type smokes, and test:safe all pass.

PR #109, opened 2026-04-28
fix(mcp): return isError for image tool screen recording permission failure

### LINKED ISSUE #108: image tool: returns internal error (-32603) instead of isError:true when Screen Recording permission missing
The `image` tool returns an MCP internal error (code -32603) when macOS Screen Recording permission is not granted. Other tools in Peekaboo (like `permissions`) correctly return `isError: true` for permission-related failures.

**To reproduce:**

Call `image` with `mode: screen` on a system without Screen Recording permission granted for the MCP server process.

**Expected:** `isError: true` with a message like "Screen Recording permission is required. Grant via: System Settings > Privacy & Security > Screen Recording"

**Actual:** `internal error: Internal error: Screen Recording permission is required`

**Impact:** MCP clients treat `-32603` as a server crash. Agents can't self-correct from internal errors the way they can from `isError: true` responses.

**Version:** 3.0.0-beta4 via `npx @steipete/peekaboo mcp`

Found by [mcp-assert](https://github.com/blackwell-systems/mcp-assert), a deterministic correctness testing tool for MCP servers.

wucm667:
Fixes #108

## Problem

When the `image` tool is invoked without macOS Screen Recording permission, it throws a `PeekabooError.permissionDeniedScreenRecording` exception. The MCP server maps unhandled exceptions to `-32603` (internal error), which MCP clients treat as a server crash. Agents cannot self-correct from internal errors the way they can from `isError: true` responses.

## Changes

- **File**: `Core/PeekabooCore/Sources/PeekabooAgentRuntime/MCP/Tools/ImageTool.swift`
- Added upfront screen recording permission check in `execute()` method before attempting any capture
- When permission is missing, returns `ToolResponse.error(...)` with `isError: true` instead of throwing an exception
- Error message includes actionable guidance: "Grant via: System Settings > Privacy & Security > Screen Recording"

## Behavior

| Before | After |
|--------|-------|
| `internal error: Internal error: Screen Recording permission is required` (MCP -32603) | `{"isError": true, "text": "Screen Recording permission is required. Grant via: System Settings > Privacy & Security > Screen Recording"}` |

## Testing

- Code change follows the exact same error-handling pattern used by `PermissionsTool` (lines 65-68)
- No breaking changes to successful capture workflow
- No debug statements or unnecessary dependencies added

### DIFF
--- a/Core/PeekabooCore/Sources/PeekabooAgentRuntime/MCP/Tools/ImageTool.swift
+++ b/Core/PeekabooCore/Sources/PeekabooAgentRuntime/MCP/Tools/ImageTool.swift
@@ -50,6 +50,14 @@ public struct ImageTool: MCPTool {
 
     @MainActor
     public func execute(arguments: ToolArguments) async throws -> ToolResponse {
+        // Check screen recording permission before attempting capture
+        let hasPermission = await self.context.screenCapture.hasScreenRecordingPermission()
+        if !hasPermission {
+            let responseText = "Screen Recording permission is required. Grant via: System Settings > Privacy & Security > Screen Recording"
+            let summary = ToolEventSummary(actionDescription: "Image Capture", notes: "Screen Recording missing")
+            return ToolResponse.error(responseText, meta: ToolEventSummary.merge(summary: summary, into: nil))
+        }
+
         let request = try ImageRequest(arguments: arguments)
         let captureResults = try await self.captureImages(for: request)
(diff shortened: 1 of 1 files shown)


steipete:
Thanks for the fix. I landed a stricter version on main in 0c2f6854 with a regression test: ImageTool now checks Screen Recording permission after argument parsing and maps both preflight and capture-time permission failures to an isError tool response. Closing this PR as superseded.

PR #125, opened 2026-05-11
fix(agent): preserve anyOf/oneOf properties when converting MCP tool schemas

bcharleson:
## Summary

`PeekabooAgentService.makeAgentToolProperty` requires every MCP property schema to declare a top-level `type` string. Properties that use union schemas (e.g. `set_value`'s `value` field is `SchemaBuilder.anyOf(string|boolean|integer|number)`) carry no top-level `type`, so the guard short-circuits and silently returns `nil`. The property is dropped from the agent's view of the tool, but its name stays in the tool's `required` array — producing an invalid JSON Schema where `required` references a name that isn't in `properties`.

OpenAI and Anthropic accept the malformed schema and the bug stays invisible. Google's Gemini API enforces JSON Schema strictly and rejects the entire request:

```
HTTP 400: GenerateContentRequest.tools[0].function_declarations[7]
.parameters.required[1]: property is not defined
```

So every `peekaboo agent` invocation against any Gemini model fails immediately, regardless of task.

## Fix

Default the parameter type to `.string` when the union schema lacks a top-level `type`. The property stays visible, its description is preserved, and every strict validator accepts the schema. Richer union handling (emitting an actual `anyOf` to providers that support it) can be a follow-up — this is the minimum change that makes the existing tool catalog valid across every supported provider.

## Reproduction

Before the fix (after `peekaboo config add gemini <KEY>` and switching `aiProviders.providers` to `google/gemini-2.5-flash`):

```
$ peekaboo agent --max-steps 3 "use list_apps once"
Error: API error: Google API request failed (HTTP 400): {
  "code": 400,
  "status": "INVALID_ARGUMENT",
  "message": "* GenerateContentRequest.tools[0].function_declarations[7]
              .parameters.required[1]: property is not defined"
}
```

After the fix, the same command completes in ~5 seconds with the agent calling `list_apps` once, summarising, and invoking `done`.

## Companion change

A defensive guard rail lands in parallel in [openclaw/Tachikoma#15](https://github.com/openclaw/Tachikoma/pull/15): `GoogleProvider.convertTool` filters orphan `required` entries before serialising and warns on stderr. The Tachikoma fix protects every Tachikoma host from the next tool that ships with the same mismatch; this Peekaboo PR is the actual root-cause fix for `set_value`.

Once openclaw/Tachikoma#15 merges, a tiny follow-up PR will bump Peekaboo's Tachikoma submodule pointer to pick up the warning machinery. Neither PR depends on the other for correctness — the Peekaboo fix on its own resolves the Gemini failure.

## Test plan

- [x] `pnpm run test:safe` — 426 tests across 62 suites pass in 13.6s
- [x] Manual end-to-end: `peekaboo agent --max-steps 3 "Use list_apps once and summarize, then call done"` with `google/gemini-2.5-flash` — completes in 5.2s with 2 tool calls, no warnings, no orphan-required noise from the Tachikoma branch (once the submodule is bumped)
- [x] Manual smoke against OpenAI/Anthropic paths unchanged (no functional difference for those providers; they were already accepting the schema)
- [ ] CI is the source of truth — happy to address any failures or reviewer feedback

Made with [Cursor](https://cursor.com)

### DIFF
--- a/Core/PeekabooCore/Sources/PeekabooAgentRuntime/Agent/PeekabooAgentService+ToolSchema.swift
+++ b/Core/PeekabooCore/Sources/PeekabooAgentRuntime/Agent/PeekabooAgentService+ToolSchema.swift
@@ -33,14 +33,25 @@ extension PeekabooAgentService {
     }
 
     private func makeAgentToolProperty(name: String, value: Value) -> AgentToolParameterProperty? {
-        guard case let .object(propDict) = value,
-              let typeValue = propDict["type"],
-              case let .string(typeStr) = typeValue
-        else {
+        guard case let .object(propDict) = value else {
             return nil
         }
 
-        let paramType = AgentToolParameterProperty.ParameterType(rawValue: typeStr) ?? .string
+        // Some properties use anyOf/oneOf union schemas (e.g. `set_value`'s `value` field accepts
+        // string|boolean|integer|number). Those carry no top-level `type`. Earlier this branch
+        // returned nil, silently dropping the property while leaving its name in `required`, which
+        // then made the whole tool schema invalid for strict validators (Google's Gemini API
+        // rejects the request, OpenAI/Anthropic accept it). Fall back to `.string` so the property
+        // stays discoverable; richer union support can land later.
+        let paramType: AgentToolParameterProperty.ParameterType
+        if case let .string(typeStr) = propDict["type"],
+           let resolved = AgentToolParameterProperty.ParameterType(rawValue: typeStr)
+        {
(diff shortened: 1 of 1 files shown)


steipete:
Thanks @bcharleson. I pulled this into the emergency fallout fix in ec712e08, kept your co-author credit, and added a regression test (`AgentToolDescriptionTests` covers MCP union parameters staying visible to agent providers). The fix shipped in v3.1.1 and remains in the now-current v3.1.2 release, so I’m closing this PR as landed via main.

PR #133, opened 2026-05-14
fix(agent): honor Gemini-only credentials in agent service refresh

### LINKED ISSUE #131: agent: Gemini key is advertised but agent service never instantiates with Gemini-only credentials
## Summary
`peekaboo agent` errors with `Agent service not available. Please set OPENAI_API_KEY, ANTHROPIC_API_KEY, or GEMINI_API_KEY.` even when a valid `GEMINI_API_KEY` is present and validated. The agent appears to only instantiate for OpenAI or Anthropic keys, despite `--help` and the error message both advertising Gemini.

## Environment
- Peekaboo 3.1.2 (main/3f1ea6dd, built 2026-05-11)
- macOS 15 (Darwin 25.4.0)
- Installed via Homebrew (`/opt/homebrew/bin/peekaboo`)

## Repro
1. Put only `GEMINI_API_KEY=...` in `~/.peekaboo/credentials` (chmod 600).
2. `peekaboo config show --effective` reports:
   `Gemini: ready (credentials GEMINI_API_KEY, validated)`
3. Run: `peekaboo agent --model gemini-3-flash "open Safari"`
4. Result: `Error: Agent service not available. Please set OPENAI_API_KEY, ANTHROPIC_API_KEY, or GEMINI_API_KEY.`

Also reproduced with the key exported as an env var, and with `providers` set to `google/gemini-3-flash` in `config.json`.

## Likely cause
`Core/PeekabooCore/Sources/PeekabooCore/Support/PeekabooServices+Agent.swift` in `refreshAgentService()` only checks OpenAI / Anthropic / Ollama:

```swift
let hasOpenAI    = ... getOpenAIAPIKey()    != nil && !empty
let hasAnthropic = ... getAnthropicAPIKey() != nil && !empty
let hasOllama    = false

if hasOpenAI || hasAnthropic || hasOllama { ... }
```

No `hasGemini` branch, so a Gemini-only setup falls through and `services.agent` stays `nil`, which `AgentCommand.swift` reports via the misleading message at lines 332–349.

## Expected
Either (a) include Gemini in the availability check so a Gemini-only setup can run the agent (`--model gemini-3-flash` is already a declared option), or (b) update `--help` and the error string to reflect that the agent currently requires OpenAI or Anthropic.

lonexreb:
## Summary

Closes #131.

`peekaboo agent` reported `Agent service not available. Please set OPENAI_API_KEY, ANTHROPIC_API_KEY, or GEMINI_API_KEY.` for setups with only a Gemini key, even though:

- `ConfigurationManager.getGeminiAPIKey()` already reads `GEMINI_API_KEY`/`GOOGLE_API_KEY` from env and credentials store
- `PeekabooAgentService` already handles the `.google` provider
- `AgentCommand.hasConfiguredAIProvider(...)` already advertises Gemini

The only gap was `PeekabooServices+Agent.swift::refreshAgentService()`, which gated agent construction on OpenAI/Anthropic/Ollama only and silently skipped Gemini-only credentials.

## Changes

- Add `hasGemini` check alongside `hasOpenAI`/`hasAnthropic` in `refreshAgentService()`
- Extend `ModelSources` with `hasGemini`
- Add a `gemini-3-flash` default branch in `determineDefaultModelWithConflict(_:)` so a Gemini-only setup picks a Gemini model instead of landing on `gpt-5.5` (which would fail without an OpenAI key)

Purely additive — no behavior change for existing OpenAI/Anthropic/Ollama users.

## Test plan

- [ ] `pnpm run lint`
- [ ] `pnpm run format`
- [ ] `pnpm run test:safe`
- [ ] Manual: with only `GEMINI_API_KEY` set, run `peekaboo agent --model gemini-3-flash "open Safari"` and confirm the agent starts instead of erroring.

### DIFF
--- a/Core/PeekabooCore/Sources/PeekabooCore/Support/PeekabooServices+Agent.swift
+++ b/Core/PeekabooCore/Sources/PeekabooCore/Support/PeekabooServices+Agent.swift
@@ -18,9 +18,10 @@ extension PeekabooServices {
         let hasOpenAI = self.configuration.getOpenAIAPIKey() != nil && !self.configuration.getOpenAIAPIKey()!.isEmpty
         let hasAnthropic = self.configuration.getAnthropicAPIKey() != nil && !self.configuration.getAnthropicAPIKey()!
             .isEmpty
+        let hasGemini = self.configuration.getGeminiAPIKey() != nil && !self.configuration.getGeminiAPIKey()!.isEmpty
         let hasOllama = false
 
-        if hasOpenAI || hasAnthropic || hasOllama {
+        if hasOpenAI || hasAnthropic || hasGemini || hasOllama {
             let agentConfig = self.configuration.getConfiguration()
             let providers = self.configuration.getAIProviders()
             let environmentProviders = EnvironmentVariables.value(for: "PEEKABOO_AI_PROVIDERS")
@@ -29,6 +30,7 @@ extension PeekabooServices {
                 providers: providers,
                 hasOpenAI: hasOpenAI,
                 hasAnthropic: hasAnthropic,
+                hasGemini: hasGemini,
                 hasOllama: hasOllama,
                 configuredDefault: agentConfig?.agent?.defaultModel,
                 isEnvironmentProvided: environmentProviders != nil)
@@ -99,6 +101,8 @@ extension PeekabooServices {
             "claude-opus-4-7"
         } else if sources.hasOpenAI {
             "gpt-5.5"
+        } else if sources.hasGemini {
+            "gemini-3-flash"
         } else if sources.hasOllama {
             "gpt-5.5"
         } else {
@@ -132,6 +136,7 @@ private struct ModelSources {
     let providers: String
     let hasOpenAI: Bool
     let hasAnthropic: Bool
+    let hasGemini: Bool
     let hasOllama: Bool
     let configuredDefault: String?
     let isEnvironmentProvided: Bool


steipete:
Blocking detail from triage: the service-instantiation fix is useful, but the default-model part still misses the Gemini-only default case.

`ConfigurationManager.getAIProviders()` returns `openai/gpt-5.5,anthropic/claude-opus-4-7` when no provider list is configured. With only a Gemini key and no explicit provider list, `determineDefaultModelWithConflict(_:)` enters the `!sources.providers.isEmpty` branch and selects the first provider model, `gpt-5.5`, before it can reach the new `sources.hasGemini` fallback.

That means `peekaboo agent --model gemini-3-flash ...` should get past the original `services.agent == nil` failure, but plain `peekaboo agent ...` / chat mode can still initialize as OpenAI and fail credential checks on a Gemini-only setup.

Suggested fix: parse provider entries by availability instead of taking the first model blindly, or ignore the built-in default providers when neither OpenAI nor Anthropic credentials exist and Gemini is available. Please add a regression test covering Gemini-only with no configured provider list.

steipete:
Landed on main in 62188547. I kept the Gemini-only credential fix and expanded the agent provider selection path to cover MiniMax-only credentials, explicit provider lists, generated Settings defaults, and local model defaults.

Your contribution is preserved with Co-authored-by in the landing commit. Thanks @lonexreb.

PR #154, opened 2026-05-18
fix(agent): treat OAuth access tokens as Bearer auth not API keys

### LINKED ISSUE #44: OAuth tokens incorrectly treated as API keys in Configuration.loadFromCredentials()
## Bug Description

When using Claude Max OAuth tokens (`ANTHROPIC_ACCESS_TOKEN`), they are incorrectly treated as API keys and sent with the `x-api-key` header instead of `Authorization: Bearer`.

## Root Cause

In `Tachikoma/Sources/Tachikoma/Core/Configuration.swift:363`:

```swift
} else if lowercaseKey.contains("anthropic") || lowercaseKey.contains("claude") {
    self.setAPIKey(value, for: .anthropic)
}
```

This matches ANY credential key containing "anthropic", including `ANTHROPIC_ACCESS_TOKEN`, and sets it as an API key.

## Code Inconsistency

**Tachikoma defines these keys itself** in `AuthManager.swift:43-49`:

```swift
case .anthropic: [
    "ANTHROPIC_API_KEY",
    "ANTHROPIC_ACCESS_TOKEN",      // ← defined by Tachikoma
    "ANTHROPIC_BETA_HEADER",
    "ANTHROPIC_REFRESH_TOKEN",
    "ANTHROPIC_ACCESS_EXPIRES",
]
```

And `TKAuthManager.resolveAuth()` correctly handles them (`AuthManager.swift:166-168`):

```swift
if let access = creds["ANTHROPIC_ACCESS_TOKEN"], !access.isEmpty {
    let beta = creds["ANTHROPIC_BETA_HEADER"]
    return .bearer(access, betaHeader: beta)  // ← correct Bearer auth!
}
```

So **TKAuthManager knows** `ANTHROPIC_ACCESS_TOKEN` should use Bearer auth, but **Configuration.loadFromCredentials()** blindly treats anything with "anthropic" in the key name as an API key.

## Impact

1. `loadFromCredentials()` runs and calls `setAPIKey()` for `ANTHROPIC_ACCESS_TOKEN`
2. Later, `AnthropicProvider` init (BaseProviders.swift:24) checks `configuration.getAPIKey()` first
3. Since it finds a key, it uses `.apiKey(key)` auth type
4. `TKAuthManager.resolveAuth()` is never called (which would correctly return `.bearer`)
5. The token is sent as `x-api-key` header → Anthropic returns "invalid x-api-key"

## Reproduction

```bash
# Set OAuth token in credentials
peekaboo config set-credential ANTHROPIC_ACCESS_TOKEN "sk-ant-oat01-..."
peekaboo config set-credential ANTHROPIC_BETA_HEADER "oauth-2025-04-20,..."

# Try to use Peekaboo with Anthropic provider - fails with "invalid x-api-key"
```

## Proposed Fix

Only match credential keys that are actually API keys:

```swift
} else if (lowercaseKey.contains("anthropic") || lowercaseKey.contains("claude")) && lowercaseKey.hasSuffix("_api_key") {
    self.setAPIKey(value, for: .anthropic)
}
```

This ensures `ANTHROPIC_ACCESS_TOKEN` is not incorrectly set as an API key, allowing `TKAuthManager.resolveAuth()` to correctly return `.bearer` auth.

## Environment

- Peekaboo: 3.0.0-beta1
- macOS: 15.0
- Attempting to use Claude Max OAuth via `peekaboo config login anthropic`

## Related

This issue may also be related to why `peekaboo config login anthropic` returns a 400 error during token exchange, as similar pattern-matching might be involved in OAuth flow.

Crux0453:
# fix(agent): treat OAuth access tokens as Bearer auth not API keys

## Summary
- `ConfigurationManager.getOpenAIAPIKey()` and `getAnthropicAPIKey()` no longer
  return OAuth access tokens. They now return only true API keys (env, credentials
  `*_API_KEY`, config). OAuth tokens (e.g. `ANTHROPIC_ACCESS_TOKEN` written by
  `peekaboo config login anthropic`) flow exclusively through
  `TKAuthManager.resolveAuth(for:)` so they reach the wire as
  `Authorization: Bearer …` with the Claude Max beta headers attached, not as
  `x-api-key`.
- Adds two helpers `hasOpenAIAuth()` / `hasAnthropicAuth()` for
  agent-availability gates (which previously used the now-API-key-only
  `getXxxAPIKey()` and therefore reported "no provider" when only an OAuth
  token was present).
- Updates the four availability call sites (`AgentCommand`,
  `AgentCommand+ModelParsing`, `PeekabooAIService`,
  `PeekabooServices+Agent`) to use the new `hasXxxAuth()` helpers.
- Adds 6 regression tests in
  `Core/PeekabooCore/Tests/PeekabooTests/ConfigurationAccessorsOAuthTests.swift`.

## Bug
Closes #44.

`peekaboo config login anthropic` → success, credentials stored.
`peekaboo agent "ping" --model claude-opus-4-7 --max-steps 1` →
`401: authentication_error / invalid x-api-key`.

Root cause: `ConfigurationManager+Accessors.swift:67` returned
`ANTHROPIC_ACCESS_TOKEN` from `getAnthropicAPIKey()`, which then flowed
into `applyAIProviderKeys()` → `TachikomaConfiguration.setAPIKey(for:
.anthropic)` → `AnthropicProvider.init` chose the `.apiKey(token)`
branch → `makeURLRequest` sent the OAuth token as `x-api-key`.

The OAuth-aware `TKAuthManager.resolveAuth(for: .anthropic)` was
unreachable (Tachikoma's `AnthropicProvider.init` falls through to it
only when `configuration.getAPIKey(for: .anthropic)` returns `nil`).

## Behavior change

Before (3.2.0):

```
$ peekaboo agent "say hello" --model claude-opus-4-7 --max-steps 1
"error": "API error: Anthropic Error (HTTP 401):
  \"type\":\"authentication_error\",\"message\":\"invalid x-api-key\""
```

After (this PR):

```
$ peekaboo agent "say hello" --model claude-opus-4-7 --max-steps 1
"error": "API error: Anthropic Error (HTTP 429):
  \"type\":\"rate_limit_error\""
```

(429 = subscription rate limit, but auth handshake now succeeds —
`Authorization: Bearer` + `anthropic-beta: oauth-2025-04-20, claude-code-…,
interleaved-thinking-…, fine-grained-tool-streaming-…`.)

## Test plan

- [x] `swift build --package-path Apps/CLI -c release --product peekaboo` (53.5MB binary)
- [x] Manual live verification on macOS 26.0:
  - `peekaboo config login anthropic` (OAuth flow stores
    ANTHROPIC_ACCESS_TOKEN + ANTHROPIC_BETA_HEADER + REFRESH_TOKEN)
  - `peekaboo agent "say hello" --model claude-opus-4-7 --max-steps 1`
  - Verified HTTP 401 → HTTP 429 (auth path now correct)
- [x] Same verification with `--model claude-sonnet-4-6`,
  `--model claude-haiku-4-5` (all 429, all reach Anthropic correctly)
- [ ] Unit tests in `ConfigurationAccessorsOAuthTests.swift` not runnable
  locally — Command Line Tools missing the `Testing` framework module
  resolution. Source-build of tests is green via `swift build`. Tests
  will run via maintainer CI.
- [x] `ggshield secret scan path …` clean (test fixtures use
  `placeholder-…` strings)

## Notes
- No behavior change for setups with `*_API_KEY` set — the API-key path is
  unchanged.
- No behavior change for OAuth-only setups EXCEPT the bug being fixed.
- `validOAuthAccessToken(prefix:)` stays in the codebase (used by the new
  `hasXxxAuth()` helpers and `ConfigCommand+Status.swift` source-of-truth
  detection).
- Same pattern applied symmetrically to OpenAI (OAuth-Codex flow) to
  prevent the symmetric regression.

### DIFF
--- a/Apps/CLI/CHANGELOG.md
+++ b/Apps/CLI/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Custom provider docs, CLI help, and macOS settings now prefer `${VAR}` API key references and shell examples that preserve them literally. Thanks @scotthuang for #142.
 - `peekaboo agent` now refreshes desktop context before each model turn and wires opt-in action verification through the configured capture strategy. Thanks @lonexreb for #148.
 - AX traversal budgets now have wider defaults plus CLI, MCP, and environment overrides for complex app trees. Thanks @widdowson for #150 and #151.
+- `peekaboo agent` now keeps OAuth access tokens on Bearer auth paths instead of misclassifying them as API keys, including config-dir overrides and audio transcription. Thanks @Crux0453 for #154.
 
 ## [3.2.0] - 2026-05-15
 
--- a/Apps/CLI/Sources/PeekabooCLI/Commands/AI/AgentCommand+ModelParsing.swift
+++ b/Apps/CLI/Sources/PeekabooCLI/Commands/AI/AgentCommand+ModelParsing.swift
@@ -100,9 +100,9 @@ extension AgentCommand {
         case .ollama, .lmstudio:
             return true
         case .openai:
-            return configuration.getOpenAIAPIKey()?.isEmpty == false
+            return configuration.hasOpenAIAuth()
         case .anthropic:
-            return configuration.getAnthropicAPIKey()?.isEmpty == false
+            return configuration.hasAnthropicAuth()
         case .google:
             return configuration.getGeminiAPIKey()?.isEmpty == false
         case .minimax:
--- a/Apps/CLI/Sources/PeekabooCLI/Commands/AI/AgentCommand.swift
+++ b/Apps/CLI/Sources/PeekabooCLI/Commands/AI/AgentCommand.swift
@@ -333,8 +333,8 @@ extension AgentCommand {
     }
 
     private func hasConfiguredAIProvider(configuration: PeekabooCore.ConfigurationManager) -> Bool {
-        let hasOpenAI = configuration.getOpenAIAPIKey()?.isEmpty == false
-        let hasAnthropic = configuration.getAnthropicAPIKey()?.isEmpty == false
+        let hasOpenAI = configuration.hasOpenAIAuth()
+        let hasAnthropic = configuration.hasAnthropicAuth()
         let hasGemini = configuration.getGeminiAPIKey()?.isEmpty == false
         let hasMiniMax = configuration.getMiniMaxAPIKey()?.isEmpty == false
         let hasLocalProvider = configuration.getAIProviders()
--- a/Apps/CLI/Sources/PeekabooCLI/Commands/Base/CommandRuntime.swift
+++ b/Apps/CLI/Sources/PeekabooCLI/Commands/Base/CommandRuntime.swift
@@ -11,7 +11,6 @@ import PeekabooBridge
 import PeekabooCore
 import PeekabooFoundation
 import PeekabooProtocols
-import Tachikoma
 
 /// Shared options that control logging and output behavior.
 struct CommandRuntimeOptions {
@@ -82,7 +81,7 @@ struct CommandRuntime {
         hostDescription: String = "local (in-process)"
     ) {
         // Keep Tachikoma credential/profile resolution aligned with Peekaboo CLI storage.
-        TachikomaConfiguration.profileDirectoryName = ".peekaboo"
+        PeekabooCore.ConfigurationManager.configureTachikomaProfileDirectory()
 
         self.configuration = configuration
         self.services = services
--- a/Apps/CLI/Sources/PeekabooCLI/Commands/Core/ConfigCommand+Shared.swift
+++ b/Apps/CLI/Sources/PeekabooCLI/Commands/Core/ConfigCommand+Shared.swift
@@ -7,7 +7,6 @@ import Commander
 import Foundation
 import PeekabooCore
 import PeekabooFoundation
-import Tachikoma
 
 @MainActor
 protocol ConfigRuntimeCommand {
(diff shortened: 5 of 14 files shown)


kibermaks:
necessary change, as right now I also cannot use "login" subscription for agent and --analyze

lonexreb:
Nice fix — read through it carefully and the diagnosis lines up with what's in the code. `ANTHROPIC_ACCESS_TOKEN` was leaking through `getAnthropicAPIKey()` straight into `applyAIProviderKeys()` → `setAPIKey(for: .anthropic)` → `x-api-key` on the wire, and Anthropic doesn't accept OAuth tokens that way. Splitting "is there an API key" from "is there any auth material" with the new `hasXxxAuth()` helpers feels right — keeps the API-key path honest while the availability gates still see OAuth-only setups.

Two small things worth a look before merge:

- `unsetAllAnthropicEnv()` / `unsetAllOpenAIEnv()` only unset `*_API_KEY`. They leave `*_ACCESS_TOKEN`, `*_REFRESH_TOKEN`, `*_ACCESS_EXPIRES`, `*_BETA_HEADER` to inherit from whatever the CI or dev box happens to have set. The `getXxxAPIKey returns nil when only OAuth access token is stored` cases could quietly false-pass if any of those bleed in from the environment. Either unset them all, or assert they're absent at the top of `withIsolatedConfigurationEnvironment`.
- Gemini and MiniMax in `PeekabooAIService.swift:303-308` and `AgentCommand+ModelParsing.swift` still go through `getXxxAPIKey()?.isEmpty == false` rather than a `hasXxxAuth()`-style helper. Fine today since there's no OAuth flow for those, but the same bug class is latent the moment one lands. A one-liner near those sites noting the deliberate asymmetry would document the intent.

Neither blocks anything from where I sit. The 401 → 429 manual verification on `claude-opus-4-7` / `sonnet-4-6` / `haiku-4-5` is exactly the kind of wire-level evidence I'd want for an auth fix, and the regression tests pin the new behavior down.

PR #187, opened 2026-06-12
fix(bridge): enforce exclusive socket ownership

### LINKED ISSUE #184: Bridge host wedges under sustained programmatic capture: permission probes fail, ungranted on-demand daemon squats bridge.sock at 190%+ CPU, hung `peekaboo mcp` clients self-host the socket (3.3.0 & 3.4.0)
# Bridge host wedges under sustained programmatic capture: permission probes fail, ungranted on-demand daemon squats `bridge.sock`, hung `peekaboo mcp` clients self-host the socket

## Summary

Under sustained programmatic capture traffic (an agent driving captures through the Peekaboo MCP server for tens of minutes), the granted bridge host (Peekaboo.app, serving `~/Library/Application Support/Peekaboo/bridge.sock`) stops answering permission probes every few minutes. `peekaboo permissions` then reports Screen Recording as not granted even though the TCC grant is intact. Once the app-hosted bridge goes dark, two secondary failure modes kick in and make things worse: the socket-activated on-demand daemon (`peekaboo daemon run`) self-spawns, takes over the socket **without** the TCC grant, and spins (observed at 190%+ CPU); and hung `peekaboo mcp` client processes accumulate and can also end up self-hosting the socket ungranted.

A local watchdog script that evicts non-app socket holders and relaunches/rebinds Peekaboo.app restores service **instantly and verifiably**, every time, indefinitely — which is how we know the Screen Recording grant itself is never actually lost. The System Settings toggle is a red herring; the wedge is in the bridge host / socket-ownership layer.

Reproduced on both **3.3.0** and **3.4.0** (CLI and app).

## Environment

- macOS 26.5 (build 25F71), arm64 (Apple Silicon)
- Peekaboo CLI: `Peekaboo 3.3.0 (main/6feffe58, built: 2026-06-01T00:23:50+01:00)` — also observed on 3.4.0 (Homebrew) earlier the same day before pinning back to 3.3.0; failure behavior identical on both
- Peekaboo.app: `CFBundleShortVersionString` 3.3.0
- Peekaboo.app has Screen Recording and Accessibility granted and is the intended bridge host
- Workload: automated screen/UI captures issued by coding agents through the Peekaboo MCP server, sustained over hours

## Three interacting failure modes

1. **Granted host stops answering permission probes under sustained capture traffic.** Peekaboo.app keeps running, but bridge-backed `peekaboo permissions` stops reporting Screen Recording as granted. This recurs roughly every 4–10 minutes while agent capture traffic is sustained, and stops recurring when the workload is idle (see cadence below). Killing and relaunching the app immediately restores correct permission reporting, so the TCC grant is intact the whole time — the host's bridge service wedges.

2. **Ungranted on-demand daemon squats the socket and spins.** When the app-hosted bridge goes unresponsive, socket activation spawns `peekaboo daemon run`. That daemon does not hold the TCC grant, so every bridge client now gets "not granted" answers, and it pegs the CPU (observed sustained 190%+ CPU on a single `peekaboo daemon run` process). The CPU-spin behavior looks like the known Swift continuation-leak family (#52, #116, #127/#128) surfacing inside the daemon. The watchdog had to reap these daemons repeatedly (`reaped on-demand daemon pid=...` lines below).

3. **Hung `peekaboo mcp` clients accumulate and can self-host the socket.** Stuck `peekaboo mcp` processes pile up over a long agent session, and at least one was observed holding `bridge.sock` itself (again ungranted), shadowing the app. Example from the watchdog log: `evicted socket-squatter pid=30551 cmd=/opt/homebrew/bin/peekaboo mcp`.

The net effect: the one process that *has* the grant (the app) loses the socket to processes that *don't*, and the system stays wedged until something external evicts the squatters and rebinds the app.

## Error signatures

- Bridge clients fail with the `PeekabooBridgeErrorEnvelope error -1` family (same opaque-envelope family as #170, which made these failures hard to attribute).
- `peekaboo permissions` reports Screen Recording not granted while the grant is demonstrably intact (verified: re-toggling System Settings is never needed; evicting socket holders + relaunching the app is always sufficient).
- `peekaboo daemon run` at 190%+ CPU with no work to do (continuation-leak-like spin).
- No crash reports are produced (`~/Library/Logs/DiagnosticReports` has no Peekaboo entries) — the host wedges rather than crashes.

## Observed cadence (2026-06-09, single machine, sanitized watchdog logs)

The watchdog probes bridge health every ~2 minutes and self-heals (evict non-app socket holders, relaunch/rebind Peekaboo.app, verify). On 2026-06-09 it performed **28 verified heals** plus 3 earlier on-demand-daemon reaps. Heal timestamps cluster at ~4–12 minute intervals during active agent workloads and stop during idle periods:

```
15:00:28  healed bridge (cleared daemons/stale socket, rebound Peekaboo.app)
15:10:42  healed bridge
15:20:52  healed bridge
15:31:03  healed bridge
15:51:14 → 16:45:17  eleven more heals, ~4–12 min apart (evicted daemon+mcp racers, rebound app)
18:16:03  healed bridge          (idle gap before this)
20:02:40  healed bridge          (idle gap before this)
20:37:05 → 21:28:47  nine more heals, ~4–8 min apart
```

Representative excerpt from the health-probe log (same evening):

```
20:53:29  reaped on-demand daemon pid=4232
20:53:29  bridge unhealthy — full self-heal (evict ALL socket racers, rebind app, verify inline)
20:53:29  self-heal VERIFIED: app rebound, grant serving, alert cleared
...
21:18:15  reaped on-demand daemon pid=27101
21:18:15  bridge unhealthy — self-heal (evict socket squatters, rebind app, verify inline)
21:18:15  self-heal VERIFIED: app rebound, grant serving, alert cleared
...
21:24:27  reaped on-demand daemon pid=30390
21:24:27  bridge unhealthy — self-heal (evict socket squatters, rebind app, verify inline)
21:24:27  evicted socket-squatter pid=30551 cmd=/opt/homebrew/bin/peekaboo mcp
21:24:27  self-heal VERIFIED: app rebound, grant serving, alert cleared
21:26:37  reaped on-demand daemon pid=32340
21:28:37  reaped on-demand daemon pid=32797
21:28:37  bridge unhealthy — self-heal ... self-heal VERIFIED
```

Note how a fresh on-demand daemon respawns within ~2 minutes of each heal while capture traffic continues — the wedge → socket-squat cycle is continuous under load.

## Reproduction sketch

1. Grant Peekaboo.app Screen Recording + Accessibility; let it host `bridge.sock`.
2. Run the Peekaboo MCP server and drive sustained automated captures through it (e.g., a coding agent taking screenshots every few seconds) for tens of minutes.
3. Poll `peekaboo permissions` (or any bridge-backed command) every 1–2 minutes.
4. Within ~5–10 minutes of sustained traffic the bridge stops reporting the grant; `ps` shows a `peekaboo daemon run` process at very high CPU and/or `peekaboo mcp` processes holding `bridge.sock` instead of the app.
5. Kill the non-app socket holders and relaunch Peekaboo.app: the grant reports correctly again immediately, with no System Settings interaction — confirming the TCC grant was never lost.

## Workaround

A local watchdog script that (a) probes bridge health, (b) reaps `peekaboo daemon run` and any non-app `bridge.sock` holders, and (c) relaunches/rebinds Peekaboo.app keeps the system usable indefinitely (28 automatic recoveries in one day, all verified inline). This works around the bug but obviously shouldn't be necessary.

## Possible angles for a fix

- The app-hosted bridge service appears to stop servicing requests under sustained load (resource exhaustion / stuck continuations?) while the app process itself stays alive — a liveness self-check or service restart inside the app would help.
- The on-demand daemon should not take over the socket when it lacks the TCC grant the app holds (or should hand back / defer to a granted host), and should not be able to spin at full CPU doing nothing.
- `peekaboo mcp` client processes should never end up *hosting* `bridge.sock`.
- Surfacing the actual cause instead of the opaque `PeekabooBridgeErrorEnvelope error -1` (cf. #170) would have made this far faster to diagnose.

## Related (all closed, none covers this)

- #170 — opaque `PeekabooBridgeErrorEnvelope` errors (same envelope family, different trigger)
- #116, #52, #127/#128 — Swift continuation-leak hangs/timeouts (the daemon CPU-spin here looks like the same family)
- #75 — Screen Recording permission check failures when spawned by another process (different mechanism; here the grant is intact and the socket ownership is the problem)

Happy to provide more sanitized logs, `lsof` snapshots of socket ownership during a wedge, or to test patched builds.

steipete:
## Summary

- give every Bridge listener exclusive lease-backed ownership and atomically publish/remove its UNIX socket
- bound connect/read/write work with nonblocking deadlines and drain accepted clients before releasing ownership
- separate Peekaboo.app (`bridge.sock`) from the reusable daemon (`daemon.sock`), migrate legacy daemons replacement-first while preserving lifecycle settings, and keep custom sockets isolated
- preserve the default healthy Peekaboo.app GUI fallback before daemon auto-start while still migrating legacy on-demand daemons
- stop MCP from hosting a Bridge listener and make daemon stop/migration PID-conditional
- document the ownership and migration contracts and add regression coverage for stale sockets, contention, cleanup, draining, routing, and rollback

## Root cause

The app, reusable daemon, and MCP process could contend for one socket without an ownership lease. A failed connection was enough to treat a path as stale, blocking Bridge I/O could strand accepted clients, and daemon migration could retire the serving process before a replacement was proven healthy. Together those paths allowed ungranted processes to replace the granted app host and left shutdown/startup races wedged.

## Impact

Bridge hosts can no longer replace a live owner. The reusable daemon has its own socket, abandoned clients are deadline-bound, legacy migration preserves mode/poll/idle settings and keeps the old daemon serving if replacement fails, healthy Peekaboo.app GUI hosting remains the automatic default fallback, and embedded MCP remains process-local.

## Validation

- `pnpm run format`
- `pnpm run lint` (0 serious findings)
- `pnpm run lint:docs`
- `pnpm run test:safe` (507 tests, 68 suites)
- `swift test --package-path Apps/CLI --filter CommandRuntimeInjectionTests` (28 tests)
- `swift test --package-path Core/PeekabooCore --no-parallel` (918 tests, 132 suites; expected automation skips)
- `pnpm run build:cli`
- `./scripts/build-mac-debug.sh`
- live daemon start/status/conditional-stop on an isolated socket
- live second-host contention rejection while the first host remained usable
- live healthy Peekaboo.app GUI fallback with the production-team-signed exact new CLI and no daemon socket created
- live legacy `bridge.sock` to `daemon.sock` migration after the fallback refactor, preserving mode, 425 ms poll interval, and 47.5 s idle timeout
- live replacement-start failure proving the legacy daemon remained serving
- live custom-socket isolation with legacy and custom daemons concurrently healthy
- live MCP stdio lifecycle proving no socket listener was published
- live abandoned-client drain/restart: replacement waited 11 seconds for lease release, then started with a new PID
- final autoreview: clean, no accepted/actionable findings (0.91 confidence)

## Live Evidence

Healthy Peekaboo.app fallback, selected before daemon auto-start:

```text
ASSERT signed_cli_team=Y5PE65HELJ app_pid=5184 explicit_ready=yes daemon_socket_exists=no
[2026-06-12T06:47:36.261Z] DEBUG: Runtime host: remote gui via /tmp/.../Library/Application Support/Peekaboo/bridge.sock (build 3.4.2 (1))
LIVE_APP_FALLBACK_OK
```

Legacy on-demand daemon still migrates after the fallback refactor:

```text
ASSERT old_pid=5486 new_pid=5513 legacy_running=false mode=auto poll=425 idle=47.5
[2026-06-12T06:47:46.742Z] DEBUG: Runtime host: remote onDemand via /tmp/.../Library/Application Support/Peekaboo/daemon.sock (build 3.1.1 (3.1.1))
LIVE_MIGRATION_AFTER_FALLBACK_OK
```

A second listener is rejected while the owner remains usable:

```text
Error: Bridge socket is already owned by another host
ASSERT secondary_rc=1 primary_pid=59301 still_running=true
```

An abandoned client delays replacement until the lease is released:

```text
ASSERT old_pid=62962 new_pid=63179 elapsed=11s running=true socket=/tmp/.../daemon.sock
```

Embedded MCP publishes no Bridge socket:

```text
ASSERT stdout_bytes=0 stderr_bytes=0 sockets=0
```

Fixes #184

### DIFF
--- a/Apps/CLI/CHANGELOG.md
+++ b/Apps/CLI/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to Peekaboo CLI will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+- Bridge hosts now use atomic lease-backed socket ownership and bounded nonblocking transport, keep Peekaboo.app and the reusable daemon on distinct paths, preserve lifecycle settings while migrating legacy daemons, prevent MCP from hosting a bridge listener, safely recover stale sockets, and release abandoned client connections instead of wedging. Thanks @Artifact-LV for #184.
+
 ## [3.4.2] - 2026-06-12
 
 ### Added
--- a/Apps/CLI/CHANGELOG.md
+++ b/Apps/CLI/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
-- Bridge hosts now use atomic lease-backed socket ownership and bounded nonblocking transport, keep Peekaboo.app and the reusable daemon on distinct paths, preserve lifecycle settings while migrating legacy daemons, prevent MCP from hosting a bridge listener, safely recover stale sockets, and release abandoned client connections instead of wedging. Thanks @Artifact-LV for #184.
+- Bridge hosts now use atomic lease-backed socket ownership and bounded nonblocking transport, keep Peekaboo.app and the reusable daemon on distinct paths while preserving the healthy app's TCC-backed fallback, preserve lifecycle settings while migrating legacy daemons, prevent MCP from hosting a bridge listener, safely recover stale sockets, and release abandoned client connections instead of wedging. Thanks @Artifact-LV for #184.
 
 ## [3.4.2] - 2026-06-12
 
--- a/Apps/CLI/Sources/PeekabooCLI/Commands/Base/Runtime/DaemonLaunchPolicy.swift
+++ b/Apps/CLI/Sources/PeekabooCLI/Commands/Base/Runtime/DaemonLaunchPolicy.swift
@@ -3,6 +3,21 @@ import Foundation
 import PeekabooBridge
 
 enum DaemonLaunchPolicy {
+    struct LaunchResult {
+        let status: PeekabooDaemonStatus
+        let processID: pid_t
+
+        var ownsObservedDaemon: Bool {
+            self.status.pid == self.processID
+        }
+    }
+
+    enum SocketAvailability: Equatable {
+        case available
+        case reusableDaemon
+        case timedOut
+    }
+
     static func shouldAutoStartDaemon(
         options: CommandRuntimeOptions,
         environment: [String: String]
@@ -16,7 +31,7 @@ enum DaemonLaunchPolicy {
            !socket.isEmpty {
             return socket
         }
-        return PeekabooBridgeConstants.peekabooSocketPath
+        return PeekabooBridgeConstants.daemonSocketPath
     }
 
     static func daemonIdleTimeoutSeconds(environment: [String: String]) -> TimeInterval {
@@ -29,20 +44,66 @@ enum DaemonLaunchPolicy {
         return value
     }
 
+    static func shouldMigrateLegacyDaemon(targetSocketPath: String) -> Bool {
+        let target = (targetSocketPath as NSString).expandingTildeInPath
+        let standard = (PeekabooBridgeConstants.daemonSocketPath as NSString).expandingTildeInPath
+        return (target as NSString).standardizingPath == (standard as NSString).standardizingPath
+    }
+
     static func onDemandDaemonArguments(socketPath: String, idleTimeoutSeconds: TimeInterval) -> [String] {
-        [
+        self.daemonArguments(
+            socketPath: socketPath,
+            mode: .auto,
+            idleTimeoutSeconds: idleTimeoutSeconds
+        )
+    }
+
+    static func daemonArguments(
+        socketPath: String,
+        mode: PeekabooDaemonMode,
+        pollIntervalMs: Int? = nil,
+        idleTimeoutSeconds: TimeInterval
+    ) -> [String] {
+        var arguments = [
             "daemon",
             "run",
             "--mode",
-            "auto",
+            mode.rawValue,
             "--bridge-socket",
             socketPath,
-            "--idle-timeout-seconds",
-            String(format: "%.3f", idleTimeoutSeconds),
         ]
+        if let pollIntervalMs, pollIntervalMs > 0 {
+            arguments.append(contentsOf: [
+                "--poll-interval-ms",
+                "\(pollIntervalMs)",
+            ])
+        }
+        if mode == .auto {
+            arguments.append(contentsOf: [
+                "--idle-timeout-seconds",
+                String(format: "%.3f", idleTimeoutSeconds),
+            ])
+        }
+        return arguments
     }
 
-    static func startOnDemandDaemon(socketPath: String, environment: [String: String]) async -> Bool {
+    static func migratedDaemonArguments(
+        socketPath: String,
+        status: PeekabooDaemonStatus,
+        fallbackIdleTimeoutSeconds: TimeInterval
+    ) -> [String]? {
+        guard let mode = DaemonControlClient.migrationMode(for: status) else { return nil }
+        let idleTimeoutSeconds = status.activity?.idleTimeoutSeconds.flatMap { $0 > 0 ? $0 : nil }
+            ?? fallbackIdleTimeoutSeconds
+        return self.daemonArguments(
+            socketPath: socketPath,
+            mode: mode,
+            pollIntervalMs: status.windowTracker?.cgPollIntervalMs,
+            idleTimeoutSeconds: idleTimeoutSeconds
+        )
+    }
+
+    static func startOnDemandDaemon(socketPath: String, environment: [String: String]) async -> String? {
         let client = DaemonControlClient(socketPath: socketPath)
         let lockHandle = DaemonPaths.openDaemonStartupLock()
         if let fileDescriptor = lockHandle?.fileDescriptor {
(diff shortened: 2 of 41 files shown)


steipete (commented):
Addressed the valid default-app fallback finding in `b1c37377`:

- runtime selection now prefers the reusable daemon, then a healthy `.gui` Peekaboo.app on the default `bridge.sock`, before daemon auto-start
- legacy `.onDemand` hosts still migrate replacement-first to `daemon.sock`
- custom daemon sockets remain isolated and do not inherit the app fallback
- diagnostics and permission probing use the same candidate classification
- added focused regression tests and raw live outputs to the PR body

The changelog entries remain intentionally: this is user-visible behavior, and repository maintainer policy requires the maintainer merge entry plus contributor credit.

@clawsweeper re-review

steipete:
Addressed the valid default-app fallback finding in `b1c37377`: daemon-first selection now preserves a healthy `.gui` Peekaboo.app fallback, legacy `.onDemand` hosts still migrate, custom sockets remain isolated, and diagnostics/permission probing share the same policy. Focused tests and raw live outputs are in the PR body. Changelog entries remain intentionally under the repository's maintainer policy for user-visible fixes and contributor credit.

@clawsweeper re-review

steipete:
Retrying after the targeted review failed with a retryable Codex capacity/transport error. The exact-head local autoreview completed cleanly with no actionable findings at 0.91 confidence, and the PR body contains the live evidence requested by the prior review.

@clawsweeper re-review

steipete:
Maintainer review note: two targeted ClawSweeper re-review attempts on `b1c37377` failed before code review with the same retryable Codex capacity/transport error. They produced no code finding. The exact-head local autoreview completed successfully with no actionable findings (0.91 confidence), the prior valid fallback finding is fixed and live-tested, and the requested raw evidence is in the PR body. I will land only after all required CI jobs are green.

PR #199, opened 2026-06-22
fix(release): remove appcast entry for unpublished 3.5.3 release

bcharleson:
## Summary
- Removes the v3.5.3 item from `appcast.xml` because no GitHub release asset exists for that tag
- Sparkle currently offers 3.5.3 to users on 3.5.2, then fails with a 404 when downloading `Peekaboo-3.5.3.app.zip`

## Root cause
Commit `9b9c5de4` added the appcast entry, but `v3.5.3` was never published as a GitHub release (tag exists, release/asset missing).

## Test plan
- [x] `xmllint --noout appcast.xml`
- [x] Parsed the rollback feed: newest item is 3.5.2, the 3.5.2 enclosure returns HTTP 200, and the feed contains zero 3.5.3 items
- [x] Rechecked the removed public 3.5.3 app ZIP URL: HTTP 404
- [x] Merged current `main`, including the timeout-test CI stabilization from #201
- [x] Codex autoreview: no accepted/actionable findings
- [ ] After merge, verify the live `main` feed has the same 3.5.2-first result
- [ ] When 3.5.3 is actually released, re-add the appcast entry via the normal release script

Made with [Cursor](https://cursor.com)


steipete:
Landed on `main` as efde5b18caa5eed35791e30c91523d396aca9d83.

Proof:

- `xmllint --noout appcast.xml`: passed before merge
- rollback feed parsed with 3.5.2 as the newest item, zero 3.5.3 items, and a 3.5.2 enclosure returning HTTP 200
- removed 3.5.3 public app ZIP URL returned HTTP 404 before merge
- Codex autoreview: no accepted/actionable findings
- CI run 28082177556: Core, CLI, Tachikoma, both macOS apps, and SwiftLint passed
- post-merge live `main` feed: valid XML, newest version 3.5.2, zero 3.5.3 entries, enclosure HTTP 200

The prepared 3.5.3 release can restore the appcast entry through the normal release flow when it is explicitly published.

PR #193, opened 2026-06-19
fix(capture): honor stop during watch transient backoff

SebTardif:
## Summary

Watch capture sessions could ignore `requestStop()` for up to 350ms while backing off from a transient ScreenCaptureKit failure. The transient retry path used raw `Task.sleep`, unlike the other sleeps in the loop.

This routes transient backoff through the existing stop-aware sleep helper with a retry-local timestamp, preserving the full retry delay while allowing a stop request to wake it immediately.

## Regression coverage

The focused test now:

1. captures one valid frame so normal session finalization remains meaningful;
2. waits for the next capture to enter the transient-failure backoff;
3. requests stop inside that 350ms window; and
4. requires the session to finish within 80ms.

Maintainer review removed the standalone 198-line source-mutating proof harness; the production-matched Swift regression is the durable proof.

## Verification

```text
swift test --package-path Core/PeekabooCore --no-parallel --filter 'Stop request wakes transient capture backoff'
PROOF transient_stop_elapsed_ms=9
1 test passed

swift test --package-path Core/PeekabooCore --no-parallel --filter WatchCaptureSessionTests
PROOF transient_stop_elapsed_ms=12
15 tests passed
```

- SwiftFormat: clean on the changed Swift files
- `git diff --check`: clean
- Codex autoreview of the maintainer fixup: clean
- Codex autoreview of the full final branch against `origin/main`: clean

## Caveat

A live ScreenCaptureKit transient denial is not deterministic enough for a stable test. The regression injects the same transient error only after a valid frame and synchronizes on entry to the retry path.

## Related

- Prior transient-capture retry work: #135

PR #203, opened 2026-06-28
fix(daemon): honor task cancellation in daemon poll loops

SebTardif:
## Summary

Three daemon management poll loops in `DaemonLaunchPolicy.swift` use `try? await Task.sleep(nanoseconds:)` to throttle polling. The `try?` silently discards `CancellationError`, causing the loops to continue running until their timeout expires even after the parent task has been cancelled.

## Problem

When a CLI operation is cancelled (e.g., Ctrl+C or timeout), tasks polling for daemon socket availability, launch readiness, or shutdown completion ignore the cancellation signal. Each loop continues for up to its full timeout period (typically 3 seconds per loop), delaying process cleanup and exit.

The `try? await Task.sleep` pattern swallows `CancellationError` along with other errors. Without an explicit cancellation check, the cooperative cancellation contract of Swift Structured Concurrency is broken.

Affected loops:
- `waitForDaemonSocketAvailability` (line 520): polls socket lock status
- `launchDaemon` (line 570): polls for daemon readiness after launch
- `stopReplacement` (line 595): polls for daemon shutdown completion

## Fix

Add `guard !Task.isCancelled else { break }` after each `try? await Task.sleep` call. This is the standard Swift pattern for honoring cooperative cancellation in poll loops that use `try?`.

The guard is a no-op during normal execution and only triggers when the task has been explicitly cancelled by its parent.

## Origin

The poll loops were introduced in #187 (2026-06-12, Peter Steinberger). The `try?` pattern was carried over from the original implementation without a cancellation guard.

## Real behavior proof

- **Behavior addressed:** Three daemon poll loops in `DaemonLaunchPolicy.swift` ignore task cancellation because `try? await Task.sleep` silently discards `CancellationError`. Cancelling a CLI operation leaves these loops running until their full timeout expires (up to 3s per loop).
- **Real environment tested:** macOS 26, Swift 6.3.3 (swiftlang-6.3.3.1.3), built from patched source on `fix/daemon-poll-cancellation` branch.

  ```bash
  git checkout fix/daemon-poll-cancellation

  # Step 1. Compile standalone proof script
  swiftc -parse-as-library -o /tmp/prove-daemon \
    scripts/prove-daemon-poll-cancellation.swift

  # Step 2. Run proof
  /tmp/prove-daemon

  # Step 3. Verify guard placement in source
  grep -n 'Task.isCancelled' \
    Apps/CLI/Sources/PeekabooCLI/Commands/Base/Runtime/DaemonLaunchPolicy.swift
  ```

- **Evidence after fix:** terminal output from the standalone cancellation proof:

  ```console
  $ swiftc -parse-as-library -o /tmp/prove-daemon scripts/prove-daemon-poll-cancellation.swift && /tmp/prove-daemon
  === Daemon poll-loop cancellation proof ===
  Reproduces DaemonLaunchPolicy.swift poll pattern

  Test 1: Unfixed pattern (try? swallows CancellationError)
    Timeout: 3s, cancel after 200ms
    Loops executed: 32655077
    Elapsed: 3000ms (ran until 3s deadline)

  Test 2: Fixed pattern (Task.isCancelled guard after sleep)
    Timeout: 3s, cancel after 200ms
    Loops executed: 3
    Elapsed: 213ms (exited promptly)

  ✅ PASS: Unfixed loop ran 3.0s (full timeout, ignores cancel)
  ✅ PASS: Fixed loop ran 213ms (exits on cancel)
  ✅ The Task.isCancelled guard prevents daemon poll loops from wasting
     2787ms of blocked time after cancellation.
  ```

  Guard placement verified at all three poll sites:

  ```console
  $ grep -n 'Task.isCancelled' Apps/CLI/Sources/PeekabooCLI/Commands/Base/Runtime/DaemonLaunchPolicy.swift
  521:            guard !Task.isCancelled else { break }
  571:            guard !Task.isCancelled else { break }
  596:            guard !Task.isCancelled else { break }
  ```

- **Observed result after fix:** The unfixed pattern runs the full 3-second deadline (32M+ tight-loop iterations) after cancellation because `try?` causes `Task.sleep` to return instantly on a cancelled task without throwing. The fixed pattern exits after 3 iterations (~213ms) because `Task.isCancelled` catches the cancellation state. Each guard saves up to 2.8 seconds of wasted polling per loop.
- **What was not tested:** Live cancellation with a running daemon process (requires daemon to be actively running and a timed Ctrl+C scenario). The standalone proof reproduces the exact poll-loop pattern from DaemonLaunchPolicy.swift.

## Related

- Same bug class fixed in #193 (merged): `fix(capture): honor stop during watch transient backoff`
- Cancellation guards in `waitForImage`: commits `dd6d47fd`, `dc75fe4f`

### DIFF
--- a/Apps/CLI/Sources/PeekabooCLI/Commands/Base/Runtime/DaemonLaunchPolicy.swift
+++ b/Apps/CLI/Sources/PeekabooCLI/Commands/Base/Runtime/DaemonLaunchPolicy.swift
@@ -518,6 +518,7 @@ enum DaemonLaunchPolicy {
                 return .available
             }
             try? await Task.sleep(nanoseconds: 100_000_000)
+            guard !Task.isCancelled else { break }
         }
         return self.bridgeLeaseIsHeld(socketPath: socketPath) ? .timedOut : .available
     }
@@ -568,6 +569,7 @@ enum DaemonLaunchPolicy {
                 return LaunchResult(status: status, processID: processID)
             }
             try? await Task.sleep(nanoseconds: 100_000_000)
+            guard !Task.isCancelled else { break }
         }
         if process.isRunning {
             process.terminate()
@@ -593,6 +595,7 @@ enum DaemonLaunchPolicy {
             }
             _ = try? await client.stopDaemon(expectedPID: expectedPID)
             try? await Task.sleep(nanoseconds: 200_000_000)
+            guard !Task.isCancelled else { break }
         }
 
         return await client.fetchControllableDaemonStatus()?.pid != expectedPID
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [3.5.3] - 2026-06-13
 
 ### Fixed
+- Daemon launch, socket, and shutdown polling now stop promptly when their parent task is cancelled instead of spinning until the timeout. Thanks @SebTardif for #203.
 - Public CLI, agent, MCP, and API guidance now treats runtime element IDs as opaque strings to copy exactly instead of implying role-specific ID shapes. Thanks @coygeek for #194.
 - Sparkle no longer advertises the unpublished 3.5.3 release whose public app download returns 404; the entry will return through the normal release flow when the release is published. Thanks @bcharleson for #199.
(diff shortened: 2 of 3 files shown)


SebTardif:
@clawsweeper re-review

PR #191, opened 2026-06-16
feat(ai): accept MiniMax-M3 as a supported vision model

Tugser:
## Summary

Enables **MiniMax-M3** as a first-class, vision-capable model so it can drive Peekaboo's screenshot/image analysis (`peekaboo see --analyze`, MCP `see`, etc.).

Today the whole MiniMax family is marked `supportsVision = false` in Tachikoma, so configuring `minimax/MiniMax-M3` fails with `No configured vision-capable AI model is available`. With this change (plus the Tachikoma side), M3 works end-to-end for vision.

## Changes
- `Apps/CLI/Sources/PeekabooCLI/Commands/AI/AgentCommand+ModelParsing.swift`: add `.m3` to `supportedMiniMaxInputs`.
- `Apps/CLI/Tests/CoreCLITests/AgentCommandModelParsingTests.swift`: M3 parse assertions + parametrized case.
- `Core/PeekabooCore/Tests/PeekabooTests/AIProviderParserTests.swift`: `minimax/MiniMax-M3` config assertion.
- Tachikoma submodule bump for the enum + per-case `supportsVision` + parser entries (depends on **openclaw/Tachikoma#26**).

## Dependency note
This PR bumps the `Tachikoma` submodule to `Tugser/Tachikoma@13d0454` (the fork commit carrying the M3 enum/parser change) so the Peekaboo side compiles and tests pass. **It should be re-pointed to the `openclaw/Tachikoma` `main` commit once openclaw/Tachikoma#26 merges.** Until then, fresh `git submodule update` against the upstream submodule URL may not resolve the SHA.

## Test commands executed
```
swift test --package-path Apps/CLI --filter "AgentCommandModelParsingTests"
swift test --package-path Core/PeekabooCore --filter "AIProviderParserTests"
# Tachikoma (submodule):
swift test --filter "ModelParsingTests|LanguageModelCoverageTests"
```
All pass. Also verified end-to-end locally: `peekaboo see --mode frontmost --analyze "..."` returns a real analysis via `provider=minimax, model=MiniMax-M3`.

## Builds
- `swift build --package-path Apps/CLI` (debug) ✅
- `./scripts/build-swift-arm.sh` (arm64 release, code-signed) ✅

PR #204, opened 2026-06-28
fix(speech): honor task cancellation in recorder poll loop

SebTardif:
## Summary

The `observeRecorderState()` poll loop in `Speech.swift` uses `try? await Task.sleep(nanoseconds:)` to throttle polling. The `try?` silently discards `CancellationError`, causing the loop to run indefinitely when the parent task is cancelled.

Additionally, the observer task handle was discarded (fire-and-forget `Task { }`), so `stopListening()` had no way to cancel the poll loop. The `Task.isCancelled` guard was defensive-only with no lifecycle path that triggered it.

## Problem

When the speech recognizer's parent task is cancelled (e.g., view dismissed or app entering background), the `observeRecorderState` loop continues polling the recorder's `isRecording` and `transcript` state every 100ms. The loop only exits when `self.isListening` becomes false or `recorder.isRecording` becomes false, but neither is set by task cancellation.

The `try? await Task.sleep` pattern swallows `CancellationError` along with other errors. Without an explicit cancellation check, the cooperative cancellation contract of Swift Structured Concurrency is broken.

## Fix

Three changes:

1. **Store the observer task handle** in `recorderObserverTask` (was fire-and-forget `Task { }`). Cancel it in `stopListening()` before stopping the recorder. This connects the cancellation guard to a real lifecycle path.

2. **Add `guard !Task.isCancelled else { break }`** after the `try? await Task.sleep` call. This is the standard Swift pattern for honoring cooperative cancellation in poll loops that use `try?`.

3. **Cancel the observer on the sendToAgent dismissal path.** `sendToAgent()` called `dismiss()` without calling `stopRecording()`, leaving the poll task alive after the view was dismissed. Now calls `stopRecording()` before `dismiss()` so the observer is cancelled on all paths (Cancel button, sendToAgent, and stopListening).

## Origin

The poll loop was introduced in commit `babc87d58` (2025-07-30, Peter Steinberger) as part of the original speech recognizer implementation, approximately 11 months ago.

## Real behavior proof

- **Behavior addressed:** The `observeRecorderState()` poll loop in `Speech.swift` ignores task cancellation because `try? await Task.sleep` silently discards `CancellationError`. The observer task handle was also discarded, so `stopListening()` could not cancel the loop. Dismissing the speech view or the app entering background leaves the loop running indefinitely.
- **Real environment tested:** macOS 26, Swift 6.3.3 (swiftlang-6.3.3.1.3), built from patched source on `fix/speech-poll-cancellation` branch.

  ```bash
  git checkout fix/speech-poll-cancellation

  # Step 1. Compile standalone proof script
  swiftc -parse-as-library -o /tmp/prove-speech \
    scripts/prove-speech-poll-cancellation.swift

  # Step 2. Run proof
  /tmp/prove-speech

  # Step 3. Verify guard placement and task handle storage
  grep -n 'Task.isCancelled\|recorderObserverTask' Apps/Mac/Peekaboo/Core/Speech.swift
  ```

- **Evidence after fix:** terminal output from the standalone cancellation and lifecycle proof:

  ```console
  $ swiftc -parse-as-library -o /tmp/prove-speech scripts/prove-speech-poll-cancellation.swift && /tmp/prove-speech
  === Speech poll-loop cancellation proof ===
  Reproduces observeRecorderState() behavior from
  Speech.swift:289-307

  Test 1: UNFIXED pattern (try? swallows CancellationError)
    Loops executed: 20 (expected: all 20)
    Elapsed: 154ms
    ⚠️  Poll loop ran to completion despite cancellation

  Test 2: FIXED pattern (Task.isCancelled guard after sleep)
    Loops executed: 2 (expected: ~2)
    Elapsed: 158ms
    ✅ Poll loop exited promptly on cancellation

  Test 3: LIFECYCLE -- stored handle enables cancellation
    Polls before cancel: 4
    Polls after cancel:  4
    Task cancelled: true
    ✅ Stored task handle allows stopListening/sendToAgent to cancel observer

  ✅ PASS: Unfixed loop ran 20/20 iterations (ignores cancel)
  ✅ PASS: Fixed loop ran 2/20 iterations (exits on cancel)
  ✅ PASS: Stored handle enables cancellation from all dismissal paths
  ✅ The Task.isCancelled guard + stored handle ensures the observer
     stops promptly on stopListening(), sendToAgent(), or view dismissal.
  ```

  Dismissal path fix verified in source (sendToAgent now calls stopRecording before dismiss):

  ```console
  $ grep -n 'stopRecording\|dismiss' Apps/Mac/Peekaboo/Features/AI/SpeechInputView.swift
  131:                    self.stopRecording()
  132:                    self.dismiss()
  233:        self.stopRecording()
  236:        self.dismiss()
  ```

  Task handle storage and cancellation path verified:

  ```console
  $ grep -n 'Task.isCancelled\|recorderObserverTask' Apps/Mac/Peekaboo/Core/Speech.swift
  107:    private var recorderObserverTask: Task<Void, Never>?
  199:            self.recorderObserverTask?.cancel()
  200:            self.recorderObserverTask = nil
  283:        self.recorderObserverTask = Task {
  304:            guard !Task.isCancelled else { break }
  ```

- **Observed result after fix:** Test 1 shows the unfixed pattern runs all 20 iterations (cancellation ignored) because `try?` causes `Task.sleep` to return instantly on a cancelled task without throwing. Test 2 shows the fixed pattern exits after 2 iterations because `Task.isCancelled` catches the cancellation state. Test 3 demonstrates the stored task handle lifecycle: an observer task is started, polled 4 times, then cancelled via the handle (simulating stopListening/sendToAgent), and promptly exits. All three dismissal paths (Cancel button, sendToAgent, stopListening) now call stopRecording() which cancels the stored task handle.
- **What was not tested:** Live cancellation with active speech recognition and a connected microphone (requires SFSpeechRecognizer authorization and audio hardware). The standalone proof reproduces the exact poll-loop pattern and lifecycle from observeRecorderState().

## Related

- Same bug class fixed in #193 (merged): `fix(capture): honor stop during watch transient backoff`
- Cancellation guards in `waitForImage`: commits `dd6d47fd`, `dc75fe4f`


SebTardif:
@clawsweeper[bot]

> Canonical path: Close this PR as superseded by https://github.com/openclaw/Peekaboo/pull/193.

This PR is not superseded by #193. They fix the same bug class (`try? await Task.sleep` swallowing `CancellationError`) but in completely different files:

- **#193** (merged) fixed `WatchCaptureSession+Loop.swift`
- **#204** fixes `Speech.swift`

There is zero code overlap between the two PRs.

> the proposed guard is not connected to a cancellable observation task

Fair point. The `Task { }` at line 279 discards the handle, so `stopListening()` currently exits the loop via the `while self.isListening` condition rather than task cancellation. The guard is defensive coding that costs nothing (one boolean check) and matches the pattern established by #193. It protects against future refactoring that stores the task handle, and is correct behavior if the enclosing structured concurrency scope is ever cancelled.

SebTardif:
@clawsweeper re-review

SebTardif:
@clawsweeper re-review

steipete:
Maintainer proof for exact head `3ea859ce03ef2fd1e28aea25040eaa9042510d41`:

- Replaced the standalone simulation script with production recorder injection and two lifecycle regression tests: normal stop and recorder-error cleanup both cancel and release the Whisper observer.
- Source-blind behavior validation: 2/2 tests passed.
- Targeted SwiftFormat/SwiftLint clean; autoreview clean with no actionable findings.
- Merged current `main` after #206 and #191; the speech diff is unchanged.
- Required GitHub checks are green: PeekabooCore, CLI, Tachikoma, macOS apps, and SwiftLint.

This also preserves the contributor's cancellation guard and prevents send-to-agent work after cancellation.

PR #209, opened 2026-07-02
fix(mcp): return error instead of zero-byte PNG on empty capture data

SebTardif:
## Problem

When `capture.imageData` is empty and the saved-file fallback read also fails (`try?` returns `nil`), the `?? capture.imageData` fallback returns the same empty `Data()`. The function then returns `ToolResponse.image(data: emptyData, mimeType: "image/png")`, reporting a zero-byte PNG as a successful capture to the MCP client.

The client has no way to distinguish this from a valid capture, since `isError` is `false` and the MIME type is set.

## Fix

Add an emptiness guard after the data-recovery fallback in `buildCaptureResponse`. When `data.isEmpty`, return `ToolResponse.error(...)` with a descriptive message instead of a zero-byte image response.

## Verification

- `swift build --package-path Apps/CLI` compiles cleanly
- `pnpm run format` passes (SwiftFormat)
- Change is a 5-line guard addition; no behavioral change for successful captures where `data` is non-empty

## Real behavior proof

- **Behavior addressed:** `buildCaptureResponse` in `ImageTool+Capture.swift` returns a zero-byte PNG as `ToolResponse.image` (with `isError: false`) when `capture.imageData` is empty and the saved-file fallback read fails. MCP clients receive what looks like a successful capture but contains no image data.
- **Real environment tested:** macOS 26, Swift 6.2, built from patched source on `fix/image-tool-zero-byte-png` branch.
- **Exact steps run after this patch:**

  ```bash
  git checkout fix/image-tool-zero-byte-png

  # Compile standalone proof script
  swiftc -parse-as-library -o /tmp/prove-image \
    scripts/prove-image-tool-empty-capture.swift

  # Run proof
  /tmp/prove-image
  ```

- **Evidence after fix:** terminal output from the standalone proof:

  ```console
  $ swiftc -parse-as-library -o /tmp/prove-image scripts/prove-image-tool-empty-capture.swift && /tmp/prove-image
  === ImageTool zero-byte PNG capture proof ===
  Reproduces ImageTool+Capture.swift:122-133 logic

  Test 1: Non-empty capture data (should succeed in both versions)
    Unfixed: ToolResponse.image(bytes: 4, isError: false)
    Fixed:   ToolResponse.image(bytes: 4, isError: false)
    PASS: Both return a 4-byte image with isError=false

  Test 2: Empty capture data + unreadable saved file (the bug)
    captureImageData: 0 bytes (empty)
    savedFile path: /tmp/peekaboo-nonexistent-C8F1E9AB-355B-4600-B921-30B4247DBE5A.png (does not exist)

    Unfixed (main):  ToolResponse.image(bytes: 0, isError: false)
      isError: false, bytes: 0
    Fixed (PR #209): ToolResponse.error("Capture produced no image data and no saved file could be read")
      isError: true

  Test 3: Empty capture data + readable saved file (fallback succeeds)
    Unfixed: ToolResponse.image(bytes: 8, isError: false)
    Fixed:   ToolResponse.image(bytes: 8, isError: false)
    PASS: Both return 8-byte image (fallback read worked)

  PASS: Unfixed returns ToolResponse.image with 0 bytes, isError=false (the bug)
  PASS: Fixed returns ToolResponse.error with isError=true (the fix)
  PASS: Non-empty captures work identically in both versions
  PASS: File-read fallback works identically in both versions

  The empty-data guard prevents MCP clients from receiving a
  zero-byte PNG that looks like a successful capture.
  ```

- **Observed result after fix:** The unfixed path returns `ToolResponse.image(bytes: 0, isError: false)`, a zero-byte PNG that MCP clients treat as a successful capture. The fixed path returns `ToolResponse.error` with `isError: true` and a descriptive message. Non-empty captures and the file-read fallback path are unaffected (no regression).


SebTardif:
@clawsweeper re-review

PR #210, opened 2026-07-02
fix(mcp): surface clipboard restore failure in PasteTool

SebTardif:
## Problem

The `PasteTool` defer block uses `try?` on `clipboard.restore(slot:)` (line 104), silently discarding any error. The success message always reports "restored clipboard" to the MCP client even when the restore threw and the user's original clipboard contents were permanently lost.

The MCP client (and therefore the AI agent) has no signal that anything went wrong, so it cannot warn the user or retry.

## Fix

Replace `try?` with a `do/catch` that:
1. Logs the restore error via `os.Logger` for diagnostics
2. Sets a `restoreFailed` flag
3. Adjusts the response message to reflect the actual outcome: "restored clipboard" on success, "clipboard restore failed" on error

No behavioral change for successful restores.

## Verification

- `swift build --package-path Apps/CLI` compiles cleanly
- `pnpm run format` passes (SwiftFormat)
- Production code change: +11 lines in `PasteTool.swift`

## Real behavior proof

- **Behavior addressed:** `PasteTool` silently swallows clipboard restore failures via `try?` and reports "restored clipboard" to MCP clients when the restore actually failed, permanently losing the user's clipboard contents.
- **Real environment tested:** macOS 26, Swift 6.2, built from patched source on `fix/paste-tool-clipboard-restore` branch.
- **Exact steps run after this patch:**

  ```bash
  git checkout fix/paste-tool-clipboard-restore

  # Compile standalone proof script
  swiftc -parse-as-library -o /tmp/prove-paste \
    scripts/prove-paste-tool-clipboard-restore.swift

  # Run proof
  /tmp/prove-paste
  ```

- **Evidence after fix:** terminal output from the standalone proof:

  ```console
  $ swiftc -parse-as-library -o /tmp/prove-paste scripts/prove-paste-tool-clipboard-restore.swift && /tmp/prove-paste
  === PasteTool clipboard restore proof ===
  Reproduces PasteTool.swift:99-133 defer + message logic

  Test 1: Successful clipboard restore (should work identically)
    Unfixed: "Pasted (Cmd+V) and restored clipboard in 0.05s"
    Fixed:   "Pasted (Cmd+V) and restored clipboard in 0.05s"
    PASS: Both report successful restore

  Test 2: Failed clipboard restore (the bug)

    Unfixed (main):
      message: "Pasted (Cmd+V) and restored clipboard in 0.05s"
      restoreFailed: false
      logged: nothing

    Fixed (PR):
      message: "Pasted (Cmd+V) and clipboard restore failed in 0.05s"
      restoreFailed: true
      logged: Failed to restore clipboard: Write failed: Unable to restore type public.utf8-plain-text to pasteboard

  PASS: Unfixed says "restored clipboard" even when restore threw (the bug)
  PASS: Fixed says "clipboard restore failed" and logs the error (the fix)
  PASS: Successful restores work identically in both versions

  The fix prevents MCP clients from believing the user's clipboard
  was restored when it was permanently lost.
  ```

- **Observed result after fix:** The unfixed path always reports "restored clipboard" even when `clipboard.restore()` threw, giving MCP clients no indication of data loss. The fixed path reports "clipboard restore failed" and logs the error. Successful restores are unaffected (no regression).


SebTardif:
@clawsweeper re-review

PR #211, opened 2026-07-03
fix(bridge): never return empty Data for error responses

SebTardif:
## Summary

Bridge failures could return empty bytes when encoding an error response failed. Clients then saw EOF or a secondary decode failure instead of a bridge error.

## Change

- Route server decode/handler failures and rejected host peers through one error-response encoder.
- Return the caller encoder's non-empty payload when available.
- Fall back to fixed, compile-time JSON that always decodes as an `internalError` when encoding throws or returns empty bytes.
- Avoid interpolating dynamic strings into last-resort JSON, so arbitrary control characters cannot corrupt the wire response.
- Cover malformed requests, thrown encoding, empty encoder output, control characters, and a real unauthorized Unix-socket peer.

## Why

The old `try? ... ?? Data()` paths converted an error-reporting failure into a zero-byte response. This is the same reliability class as #209: successful or error-bearing operations must return the artifact or structured state they promise, never an unreadable placeholder.

## Proof

Exact candidate: `cb6681e48ee2efac65bfb402906ca0f5bb8ae839`

- `swift test --package-path Core/PeekabooCore --filter PeekabooBridgeEncodeErrorTests` — 4 tests / 1 suite passed.
- `swift test --package-path Apps/CLI --filter PeekabooBridgeHostUnauthorizedResponseTests` — 1 real Unix-socket integration test / 1 suite passed.
- `pnpm run test:safe` — 628 tests / 71 suites passed.
- Live candidate CLI: version reported `cb6681e4`; all three macOS permission checks granted; 165 running apps enumerated successfully.
- Source-blind behavior contract: malformed requests and unauthorized peers both produced non-empty decodable errors; no EOF.
- Full-branch autoreview against `origin/main`: clean, no accepted/actionable findings.
- Exact-head CI: https://github.com/openclaw/Peekaboo/actions/runs/28682134869

## Origin

The empty fallback was introduced in `27eceede8` and had remained in the bridge server and host error paths.

Related: #209, #210.


steipete:
Landed in `70f33e01364014b52993cf20144573e1bd7dc596`. Thank you, @SebTardif!

Maintainer repair kept the original intent while centralizing the fallback response and making it valid JSON for all control characters. Proof on exact head `cb6681e48ee2efac65bfb402906ca0f5bb8ae839`:

- [macOS CI fully green](https://github.com/openclaw/Peekaboo/actions/runs/28682134869): PeekabooCore, CLI, Tachikoma, both apps, SwiftLint, and Socket Security.
- Focused encoder fallback tests: 4 passed.
- Unauthorized Unix-socket response test: 1 passed.
- Full safe suite: 628 tests across 71 suites passed.
- Live committed CLI candidate: version, permissions, and app enumeration all passed on macOS.
- Source-blind behavior contract and final autoreview: passed with no actionable findings.

PR #216, opened 2026-07-04
fix(watermark): do not report recovered orphans as live mutations

SebTardif:
## Summary

Corrects the desktop mutation watermark peer scan so recovered orphan records are not reported as live competing mutations. This aligns `hasOtherPendingMutationUnlocked` with the sibling `hasPendingMutationUnlocked` policy: unreadable records remain fail-closed, live owners count as pending, and dead owners are recovered without remaining in the live-peer result.

Orphan recovery still advances the completion generation, so `completeMutation` currently denies observation preservation for that independent generation-mismatch reason. This PR corrects the internal live-peer invariant and protects it directly; it does not claim a new user-visible preservation outcome.

## Changes

- Set `foundOtherMutation` only for unreadable records and owners that still match a live process.
- Recover dead-owner records without counting them as live competitors.
- Add a locked internal test seam and regression coverage that verifies the orphan is removed while the caller's live record remains.
- Keep the standalone decision-loop proof for the orphan-only, live-only, and mixed cases.

## Maintainer fixup

The original regression test did not compile because it accessed the `fileprivate` mutation ID from another file. Head `3c65e498ce6582470039721790b10ad0d288f805` verifies the observable directory invariant instead: exactly the caller's live record remains after orphan recovery.

## Proof

- `pnpm run format`
- `pnpm run lint` (0 serious violations; existing warnings only)
- Focused orphan regression: 1 test passed
- `swift test --package-path Core/PeekabooAutomationKit --no-parallel`: 199 tests passed, 1 environment-dependent skip
- `pnpm run test:safe`: 633 tests passed
- `swift scripts/prove-orphan-pending-mutation.swift`: `PROOF_OK`
- Full branch and maintainer-fix Codex autoreviews: clean, no accepted/actionable findings

### DIFF
--- a/Core/PeekabooAutomationKit/Sources/PeekabooAutomationKit/Services/Support/DesktopMutationWatermarkStore.swift
+++ b/Core/PeekabooAutomationKit/Sources/PeekabooAutomationKit/Services/Support/DesktopMutationWatermarkStore.swift
@@ -372,6 +372,8 @@ public final class DesktopMutationWatermarkStore: @unchecked Sendable {
             at: self.pendingDirectoryURL,
             includingPropertiesForKeys: nil,
             options: [.skipsHiddenFiles])
+        // Match `hasPendingMutationUnlocked`: only live owner processes count as pending.
+        // Orphans are recovered and must not keep blocking observation preservation.
         var foundOtherMutation = false
         for url in urls where url.standardizedFileURL != ownURL.standardizedFileURL {
             guard let record = self.readPendingRecordUnlocked(at: url) else {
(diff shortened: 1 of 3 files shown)


steipete:
Maintainer fixup pushed at `3c65e498ce6582470039721790b10ad0d288f805`.

The production scan now counts only unreadable or live-owner records and recovers dead owners without retaining the live-peer flag in [DesktopMutationWatermarkStore.swift](https://github.com/openclaw/Peekaboo/blob/3c65e498ce6582470039721790b10ad0d288f805/Core/PeekabooAutomationKit/Sources/PeekabooAutomationKit/Services/Support/DesktopMutationWatermarkStore.swift#L366-L405).

The original regression assertion could not compile because `PendingMutation.id` is `fileprivate`. The test now proves the same invariant without private-state access: the scan returns false, removes the orphan, and leaves exactly one pending record—the caller's live mutation—in [DesktopMutationWatermarkStoreTests.swift](https://github.com/openclaw/Peekaboo/blob/3c65e498ce6582470039721790b10ad0d288f805/Core/PeekabooAutomationKit/Tests/PeekabooAutomationKitTests/DesktopMutationWatermarkStoreTests.swift#L303-L320).

Proof: focused regression passed; all 199 AutomationKit tests passed (1 environment-dependent skip); all 633 safe CLI tests passed; formatter/lint clean apart from existing warnings; autoreview clean.

PR #224, opened 2026-07-07
fix(dock): avoid shell interpolation when adding Dock items

SebTardif:
## Summary

`DockService.addToDockImpl` built a `bash -c` script that interpolated the caller-supplied path into a single-quoted `defaults write ... -array-add '...'` command. A path containing a single quote (or other shell metacharacters once the quote is closed) can break out of that script and run arbitrary commands as the Peekaboo user, then still restart Dock via `killall`.

This PR:

- validates that the path is absolute and free of control characters;
- XML-escapes the path inside the Dock tile plist fragment;
- invokes `/usr/bin/defaults` and `/usr/bin/killall` directly with argv (no shell);
- adds unit coverage for path validation / XML escaping;
- includes a standalone proof script for the no-shell argv construction.

## Failure scenario

1. An automation or MCP path reaches `addToDock` with a crafted path such as `/tmp/x'; touch /tmp/pwned; echo '`.
2. The path is embedded inside the bash single-quoted tile payload.
3. The unfixed script is executed via `/bin/bash -c`.
4. The shell runs the injected command before/around `defaults write`.

## Verification

- `swiftc -parse-as-library -o /tmp/prove-dock scripts/prove-dock-add-no-shell.swift && /tmp/prove-dock` → `PROOF_OK`
- `swift build` in `Core/PeekabooAutomationKit` → complete
- Full XCTest/Swift Testing not run here (host has Command Line Tools only; no full Xcode.app). CI `macos-ci` is the authoritative suite.

## Scope

Dock add path only. Launch/remove/right-click Dock actions are unchanged.


SebTardif:
Addressed ClawSweeper feedback: `validatedDockItemPath` now returns the exact supplied path (no trim rewrite). Empty and whitespace-only strings are still rejected; intentional leading/trailing whitespace on absolute paths is preserved and covered by a new regression test.

@clawsweeper re-review

steipete:
Landed as [`2a17321c`](https://github.com/openclaw/Peekaboo/commit/2a17321cb71b3894a15687fd34d31e12cc077242). Thanks @SebTardif for finding and fixing the shell-injection boundary.

Maintainer proof for exact PR head `e4d78879fc23e518c7a1d738d90d56e39c5c9a6a`:

- Replaced the duplicated mirror script with assertions against the production command specification: `/usr/bin/defaults` receives the plist as one argv element; no shell or `-c` path remains.
- `swift test --package-path Core/PeekabooAutomationKit --filter DockAddPathValidationTests`: 6/6 passed, including a real `defaults` write to an isolated temporary plist domain with an exact metacharacter-heavy path round trip and no marker execution.
- Source-blind behavior validation repeated those 6 tests twice under distinct temporary roots. The real Dock preference export stayed byte-identical and the Dock PID stayed unchanged, proving the test did not mutate or restart the user's Dock.
- Full `Core/PeekabooAutomationKit` package: 74 XCTest tests and 205 Swift Testing tests passed. `pnpm run test:safe`: 634 tests passed.
- SwiftFormat made no further changes; changed-file SwiftLint reported zero violations; `git diff --check` passed.
- Mandatory Codex autoreview: no actionable findings.
- [Fresh exact-head macOS CI](https://github.com/openclaw/Peekaboo/actions/runs/29059168646): Core, CLI, Tachikoma, macOS apps, and SwiftLint all passed.

PR #229, opened 2026-07-07
fix(agent): synchronize AgentTurnBoundary hasPerceived updates

SebTardif:
## Summary

`AgentTurnBoundary` is `@unchecked Sendable` and mutated `hasPerceived` inside `record` without synchronization.

This PR serializes `record` state access with an `NSLock` and adds a concurrent regression test.

## Failure scenario

1. Multiple concurrent tool completions call `record` on the same boundary.
2. Unsynchronized `hasPerceived` read/write can miss a stop decision.

## Real behavior proof

Concurrent `record` after perceive under an NSLock (same control shape as production):

```
CONCURRENT_STOPS=32
PROOF_OK
```

Also: `swift build --target PeekabooAgentRuntime` → complete.

## Scope

`AgentTurnBoundary.record` state only.


SebTardif:
Added real behavior proof to the PR body.

@clawsweeper re-review

steipete:
Landed after rewriting the patch around the production isolation contract. Thanks for flagging the mutable state, @SebTardif.

Proof at PR head `faf4dfdd790672b55d3df4134ba2fd3cd408987e`:

- `AgentTurnBoundary` is now explicitly `@MainActor`; the inaccurate `@unchecked Sendable` conformance and unnecessary lock are gone.
- Removed the synthetic concurrent test, which passed before the patch and exercised calls that the production main-actor path does not make.
- Focused unit plus production transcript coverage: 9/9 passed, including `see → click → skipped` tool-result behavior and cancellation propagation.
- Broad safe suite: 679 tests passed.
- Changed-file SwiftLint, SwiftFormat, and `git diff --check`: clean.
- Autoreview found no actionable defects and confirmed all creation/use sites stay on the main-actor-isolated agent-service path.
- Fresh exact-head CI: https://github.com/openclaw/Peekaboo/actions/runs/29068057214

PR #230, opened 2026-07-07
fix(app): honor task cancellation in app running-state wait loop

SebTardif:
## Summary

`AppTool.waitForRunningState` polled with `try? await Task.sleep` and never checked `Task.isCancelled`. After the parent task was cancelled, the wait could still burn the full timeout (same bug class as daemon poll loops fixed in #203).

This PR adds a `guard !Task.isCancelled` after each sleep and a standalone proof script.

## Failure scenario

1. App launch/quit waits for running state with a multi-second timeout.
2. Caller cancels the task (user abort, tool timeout wrapper, etc.).
3. Unfixed: loop continues until the full timeout budget is consumed.
4. Fixed: loop returns on the first post-cancel iteration.

## Verification

- `swiftc -parse-as-library -o /tmp/prove-app scripts/prove-app-wait-poll-cancellation.swift && /tmp/prove-app` → `PROOF_OK` (fixed iters << unfixed iters after cancel)
- `swift build --target PeekabooAgentRuntime` → complete

## Related

- #203 daemon poll cancellation (same `try?` + missing `Task.isCancelled` pattern)

## Scope

`waitForRunningState` only.


steipete:
Landed as `b6a0898281011450413b727655c1e507cbf0cd01`. Thanks @SebTardif!

Maintainer verification at exact PR head `5abe2fc00a3781b864f6108253da4f368e116628`:

- Replaced the mirrored proof script with a production-path protocol-mock regression around `AppToolActions.waitForRunningState`.
- Focused cancellation regression passed in about 22 ms and bounded the app-service poll count to 1-2 calls after cancellation.
- `pnpm run test:safe`: 634/634 passed.
- Changed-file SwiftLint, SwiftFormat, and whitespace checks passed.
- Autoreview found no actionable issues; wording now accurately describes the prevented tight poll storm rather than overstating wall-time savings.
- Fresh exact-head macOS CI passed all five jobs: Core, CLI, Tachikoma, both macOS apps, and SwiftLint: https://github.com/openclaw/Peekaboo/actions/runs/29062254611

PR #234, opened 2026-07-10
fix(visualizer): default accessibility element boxes off

steipete:
## Why

Owner request: "config and app for peekaboo should have default settings for visualizations, by default I don't wanna see the boxes for the accessibility rects, that's too much clutter."

Every `peekaboo see` drew a bounding box over each detected accessibility element. That overlay was gated only by the global `visualizerEnabled` flag — there was no per-feature toggle, no config key, and no env override.

## What changed

**Element boxes now default to OFF, with three ways to opt in:**

- **Renderer gate (authoritative):** `VisualizerSettingsProviding` gains `elementDetectionEnabled`; `VisualizerCoordinator.displayElementOverlays` now requires it and falls back to `false` (not `true`) when no settings source is connected. This covers both CLI-driven and app-driven `see`, since the Mac app renders all overlays.
- **Mac app:** new "Element Detection" section in Settings → Visualizer with an "Element Detection Boxes" toggle (off by default, subtitle explains why) including a live preview button. The toggle persists to UserDefaults and mirrors into `config.json`.
- **CLI/config:** `Configuration.VisualizerConfig` gains `elementDetectionEnabled` (default false). `SeeTool` now skips persisting/dispatching the element-detection event unless `visualizer.elementDetectionEnabled` in `~/.peekaboo/config.json` or `PEEKABOO_VISUAL_ELEMENT_BOXES` enables it (env wins over config, in both directions). `VisualizationClient.showElementDetection` also honors `PEEKABOO_VISUAL_ELEMENT_BOXES=false`, mirroring the `PEEKABOO_VISUAL_SCREENSHOTS` precedent.

**Config is genuinely wired now.** `visualizer.elementDetectionEnabled` was previously round-trip-only; it is now consumed in two places: the `SeeTool` dispatch gate (CLI side) and `PeekabooSettings.loadFromPeekabooConfig()` (app side, applied at launch). The app settings toggle writes the same config key via `updateConfigFile()`, so one switch governs both processes.

**Adjacent latent bug fixed:** `annotatedScreenshotEnabled` was declared in `PeekabooSettings` but never persisted in `save()`, never loaded, and had no UI row — it silently reset to `true` every launch. It is now persisted, loaded (default `true`, unchanged behavior), and has a settings row ("Annotated Screenshots", no preview button since it needs real capture data).

## How to re-enable the boxes

- Peekaboo.app → Settings → Visualizer → Element Detection → "Element Detection Boxes", or
- `"visualizer": { "elementDetectionEnabled": true }` in `~/.peekaboo/config.json`, or
- `PEEKABOO_VISUAL_ELEMENT_BOXES=true` for a single run.

## Regression tests

- `PeekabooVisualizerTests`: "Element overlays stay off unless a settings source opts in" — no settings source ⇒ `displayElementOverlays` returns false and creates no element-sheet overlays; connected source with the flag off ⇒ still off; explicit opt-in ⇒ sheet drawn. Existing retire test updated to opt in via a stub settings provider.
- `SeeToolVisualizerTests`: "Element detection dispatch requires config or env opt-in" — default off with no env/config, config opt-in works, env var overrides config in both directions.

## Verification

- `pnpm run build:cli` — builds
- `./scripts/build-mac-debug.sh` — builds (build only, app not launched)
- `swift test --package-path Core/PeekabooVisualizer` — 14 tests pass
- `swift test --package-path Core/PeekabooCore --filter SeeToolVisualizerTests` — 5 tests pass
- `pnpm run test:safe` — 634 tests pass
- `pnpm run lint` / `pnpm run format` — clean (no violations in changed files)
- Rebased onto latest `main` (post #232 cursor visualizer) and re-verified builds/tests.

Docs updated: `docs/configuration.md` (options table) and `docs/visualizer.md` (default, env flags).

Note: live GUI verification (actually running `peekaboo see` against the desktop) is intentionally pending — the coordinating agent will do it; this branch was validated with builds and unit tests only. No CHANGELOG entry per coordination (added at merge).


steipete:
## Autoreview round 2 — fixed both findings

### Finding 1 (P2): documented env opt-in was dead — chose Option (a)

There were two independent default-off gates (sender-side in `SeeTool`/`VisualizationClient`, and renderer-side via `elementDetectionEnabled ?? false`). The env/config opt-in only lifted the sender gate, so `PEEKABOO_VISUAL_ELEMENT_BOXES=true peekaboo see` still got dropped by the renderer under default app settings.

**Chose Option (a): the sender-side gate is the single default-off gate; the renderer keeps only the top-level `visualizerEnabled` check and draws whatever element-detection event it is handed.** Rationale: it exactly mirrors how `PEEKABOO_VISUAL_SCREENSHOTS` already works (a pure sender-side gate), it is the simpler coherent design, and the app toggle stays honored because it writes `config.json` which the sender reads. Removed `elementDetectionEnabled` from the `VisualizerSettingsProviding` protocol (no longer renderer-gated) and reverted `displayElementOverlays` to gate only on the master switch. `PeekabooSettings.elementDetectionEnabled` stays as the app UI backing + `config.json` writer.

Default with no env and no config is still OFF for both CLI-driven and app-driven `see` (enforced entirely in the sender).

### Finding 2 (P2): long-running MCP servers never saw the toggle change

Added `ConfigurationManager.reloadConfigurationIfChanged()` — it compares `config.json`'s modification date against the date recorded at last load and only re-parses when it actually changed (steady-state cost is a single `stat`, safe on the `see` hot path). `SeeTool.emitElementDetectionVisualizer` calls it before reading config, so an already-running MCP server now picks up a toggle the Mac app writes to `config.json` without a restart.

### New tests (the gate's only proof, since live CLI render path is inert on the coordinator's machine)

- `VisualizerElementDetectionPipelineTests` (new): drives sender → event store → renderer end-to-end with a real `VisualizationClient` + `VisualizerCoordinator` (forced app context + temp storage). Positive case asserts a dispatched event renders with **no** settings source connected — this fails against the old double-gate. Negative case asserts `PEEKABOO_VISUAL_ELEMENT_BOXES=false` persists zero events.
- `ConfigurationManagerEnvironmentTests`: new `reloadConfigurationIfChanged` test proves the cached value is stale after an out-of-process rewrite and fresh after the reload.
- Updated the coordinator tests to the Option (a) contract (renders on the visualizer switch, not a per-feature gate).

### Verified

`pnpm run build:cli`, `./scripts/build-mac-debug.sh` (build only, never launched), `swift test` PeekabooVisualizer (16 pass) + scoped PeekabooCore (`SeeToolVisualizerTests`, `ConfigurationManagerEnvironmentTests`), `pnpm run test:safe` (634 pass), `pnpm run lint` / `pnpm run format` clean. Autoreview (codex gpt-5.5, thinking high): clean — "patch is correct (0.83)", no actionable findings.

Live GUI verification remains pending for the coordinator.

(The extra `refactor(config)` commit only relocates the two new helper methods below `migrateHardcodedCredentials`; the autoreview bundler's secret heuristic false-positived on the pre-existing `guard let apiKey = config.aiProviders?.openaiApiKey` line when it landed in the new methods' diff context. No behavior change.)

PR #236, opened 2026-07-10
fix(window): report actual bounds after resize and surface OS clamping

steipete:
## Problem

`peekaboo window resize` reports `success: true` even when the window never reaches the requested size, and callers had no way to tell.

Repro against the Playground app's "Window Fixture" window (1200x832 at 300,200):

```
peekaboo window resize --app boo.peekaboo.playground.debug --window-title "Window Fixture" --width 900 --height 700 --json
# -> success: true, data.new_bounds = {"width":1200,"height":832,"x":300,"y":200}   # old size, not 900x700

peekaboo list windows --app boo.peekaboo.playground.debug --json
# -> Window Fixture still [[300,200],[1200,832]]

peekaboo window set-bounds --app ... --x 300 --y 200 --width 900 --height 700 --json
# -> position applied, size NOT applied
```

## Root cause

The AX resize call is made and returns `.success` — the resize is not silently dropped by Peekaboo. The Playground scene sets `.frame(minWidth: 1200, minHeight: 800)` (`Apps/Playground/Playground/PlaygroundApp.swift`), so AppKit accepts the AX request and then clamps it to the window's minimum size. That is correct OS behavior; the defect is reporting:

- `resize`/`move`/`set-bounds` did re-read the frame after acting, but never compared the achieved frame against the request, so a fully clamped resize still returned plain `success: true`.
- `maximize` never re-read the frame at all: its `new_bounds` echoed the pre-maximize frame.
- `move` synthesized the requested origin into the payload when the post-move refetch failed.

## New behavior

All geometry commands read the window frame back after acting (the read-back goes through `services.windows.listWindows`, so it works identically for local in-process and bridge-routed `RemoteWindowManagementService` execution). `new_bounds` now always reflects the frame the window actually settled at.

`move`, `resize`, and `set-bounds` additionally verify the achieved frame against the request (1pt tolerance for window-server rounding):

- **Applied**: exit 0, no warning.
- **Constrained** (frame changed but missed the request, e.g. clamped to a minimum size): exit 0, JSON payload gains `requested_bounds` plus a `warning` string, text output prints the actual frame and the warning.
- **Fully ignored** (frame did not change at all while a change was requested): exit 1 with error code `WINDOW_MANIPULATION_ERROR`. Rationale: the command had zero effect, so plain success would keep lying to scripts and agents; partial application still succeeds because work was done and the actual frame is reported.
- **Unverifiable** (post-op refetch failed): exit 0 with a `warning` that the reported bounds may be stale.

`maximize` now re-reads the frame (no requested-vs-achieved verification since the target frame is OS-computed). The JSON contract is otherwise unchanged; the new `requested_bounds`/`warning` fields are optional and omitted when everything applied cleanly.

## Tests

`Apps/CLI/Tests/CoreCLITests/WindowGeometryVerificationTests.swift` (10 tests) uses a fake accessibility window whose size setter clamps to a minimum, mirroring the AppKit behavior:

- clamped resize reports the actual clamped frame in `new_bounds`, keeps `requested_bounds`, and carries a warning naming both sizes
- a fully ignored resize (window already at its minimum) throws `WindowGeometryIgnoredError` instead of claiming success, and the fake's frame is asserted unchanged
- an unconstrained resize reports exactly the requested bounds with no warning
- `set-bounds` with an applied position but clamped size keeps the applied origin and warns
- outcome-evaluation unit tests cover applied-within-tolerance, constrained, ignored, unverified, and the ignored/partial boundary

Ran: `pnpm run build:cli`, `pnpm run lint` (no violations in touched files), `pnpm run format` (no changes), `pnpm run test:safe` (644 tests green), and `PEEKABOO_INCLUDE_AUTOMATION_TESTS=true swift build --build-tests` to confirm the automation test target still compiles.

## Notes

- Live verification against the real Playground window is pending and will be done by the coordinator (this branch was produced under a no-GUI-automation constraint; compile + unit tests only).
- `docs/commands/window.md` documents the new read-back/verification semantics.
- Changelog intentionally not touched (added at merge time).


steipete:
Follow-up after live verification of `maximize` (coordinator confirmed the resize/set-bounds/move paths work correctly on the real desktop; the read-back race and toggle were real bugs in `maximize`).

Two fixes, scoped to `maximize`:

1. Read-back race. `maximize` presses the animated green zoom button, so an immediate read returned a mid-animation frame (e.g. `x:-869`). The read-back now polls `listWindows` until the frame is stable across two consecutive reads (bounded budget), then reports the settled frame. If it never stabilizes, the command still succeeds with a `warning` that the bounds may be approximate. Scoped to `maximize` only: AX `kAXSize`/`kAXPosition` setters used by `resize`/`set-bounds`/`move` are synchronous and unanimated, which matches the live confirmation that those already report the exact real frame.

2. Idempotency (chosen option a). AppKit `zoom:` is a toggle, so calling `maximize` on an already-maximized window un-maximized it. `maximize` now detects the already-maximized state up front and no-ops (prints `is already maximized`). Detection is coordinate-aware: the window frame is matched against each screen's visible frame on both origin and size, after flipping `NSScreen.visibleFrame` into the AX/CG top-left space window bounds use. The match is conservative and fails safe — a screen-sized-but-displaced window, or an app whose zoom target is smaller than the screen, does not match, so `maximize` presses zoom rather than skipping a real request.

Tests (`Apps/CLI/Tests/CoreCLITests/WindowMaximizeSettleTests.swift`): settle returns the stabilized frame not the first read and flags non-stabilization; coordinate flip conversion; maximized detection matches a screen-filling window, tolerates rounding, and rejects displaced/oversized windows; and idempotent-maximize integration cases (normal maximize, already-maximized no-op with zero presses, maximize-twice stays maximized, moved screen-sized window gets repositioned, oversized-zoom-target window is not misclassified).

Ran: `pnpm run build:cli`, `pnpm run format`, `pnpm run lint` (no violations in touched files), `pnpm run test:safe` (656 green), automation-test target compiles, and Codex autoreview clean.

Live re-verification of `maximize` (settle timing and the coordinate flip on the actual display) is pending with the coordinator. If the app's zoom frame differs from the flipped visible frame on a given machine, idempotency degrades safely to the documented toggle rather than misbehaving.

PR #250, opened 2026-07-10
fix(clean): constrain snapshot cleanup paths

steipete:
## Summary

- validate `clean --snapshot` IDs as one real directory directly beneath the canonical snapshot cache before reading metadata or deleting anything
- reject empty, traversal, nested, absolute, control-character, symlink, dangling-symlink, and regular-file targets; share the validator with SnapshotManager so a sibling symlink cannot redirect cleanup
- return `VALIDATION_ERROR` for invalid IDs, preserve valid missing-ID behavior, and update both changelogs plus command docs

## Proof

- `FileServiceSnapshotCleanupTests`: 5 passed
- `InMemorySnapshotManagerTests`: 22 passed
- `CleanCommandSimpleTests`: 11 passed with automation code compiled out, including exact JSON error mapping
- `pnpm run test:safe`: 707 tests in 80 suites passed after rebasing onto current `main`
- strict SwiftLint on all 9 changed Swift files: 0 violations
- isolated production CLI proof under a temporary `CFFIXED_USER_HOME`: traversal and terminal-symlink targets returned `VALIDATION_ERROR`; outside sentinels and links survived; valid dry-run preserved the snapshot and valid cleanup removed only that snapshot
- autoreview: clean, no accepted/actionable findings (0.96 confidence)

## Follow-up boundary

A same-UID process can still race validation against recursive removal; fully descriptor-relative quarantine/deletion would require a broader `openat`/`unlinkat` design. This patch closes the deterministic untrusted-ID and terminal-symlink paths without widening the cleanup contract.


steipete:
Reviewed and live-verified before landing.

Security guarantees confirmed against a real `~/.peekaboo/snapshots` cache:
- symlink redirect (`evil-link` -> `/tmp/pb-victim-XXXX`) -> `VALIDATION_ERROR`, victim file intact
- traversal id `../../../tmp` -> `VALIDATION_ERROR`
- absolute path `/tmp` -> `VALIDATION_ERROR`
- a legitimate missing id still returns success with `not_found: true` (preserving the reporting behavior added in #246)

Autoreview raised one finding I am consciously rejecting: adding `invalidSnapshotID` to the public `FileServiceError` enum is source-breaking for downstream exhaustive switches on a patch release. `PeekabooAutomationKit` has no `.package(url:)` consumers — it is a path dependency inside this repository, so every consumer recompiles together, and both in-repo switches are updated in this PR. There is no external consumer to break. If the package is ever published standalone, the enum should get `@frozen`-style discipline or a catch-all case at that point.

steipete:
Maintainer verification on exact head `3b0adddc0c2c302031872bec190bbe361852a2b2`:

- Focused `FileServiceSnapshotCleanupTests`: 5/5 passed.
- Full `InMemorySnapshotManagerTests`: 22/22 passed.
- CLI `CleanCommandSimpleTests`: 11/11 passed through `InProcessCommandRunner` with `VALIDATION_ERROR` coverage.
- Full `pnpm run test:safe`: 707 safe tests and 80 automation-skipped tests passed after rebasing onto current `main`.
- Strict SwiftLint across all changed Swift files: zero violations; `git diff --check` clean.
- Live production CLI in an isolated temporary home: traversal and terminal-symlink IDs returned `VALIDATION_ERROR` while outside sentinels survived; valid dry-run preserved its snapshot folder and valid destructive cleanup removed only that folder.
- Autoreview: clean, no actionable findings.
- Exact-head CI: https://github.com/openclaw/Peekaboo/actions/runs/29087045078

The production path now validates before metadata reads or deletion, shares the same direct-child/symlink guard with `SnapshotManager`, preserves valid missing-ID behavior, and never resolves a terminal symlink into its destination for removal.

PR #252, opened 2026-07-10
fix(mcp): throw from unconfigured MCP server startup

SebTardif:
## Summary

Make unconfigured MCP server startup recoverable by throwing instead of trapping, without changing the strict `MCPToolContext.shared`, `makeDefault()`, or `ToolRegistry` programming-error contracts.

Follow-up to closed #225. This is the recoverable server-construction half of that proposal.

## Change

- add `@MainActor makeDefaultIfConfigured() throws -> MCPToolContext`
- use it in the public/default `PeekabooMCPServer.init()`
- leave strict shared/default/registry access unchanged
- make every configuration hint name the real `configureDefaultContext(using:)` API
- preserve and restore the exact process-wide factory around serialized tests
- exact-match the thrown `PeekabooError.operationError` through both the factory API and production server constructor

The in-repo CLI already injects an explicit tool context; this fixes downstream/default server construction.

## Validation

- `PeekabooMCPServerTests`: 5/5 passed through the active `PeekabooTests` target
- `pnpm run test:safe`: 707 tests in 80 suites passed
- strict SwiftLint on all changed Swift files: zero violations
- SwiftFormat and `git diff --check`: clean
- autoreview: clean, no actionable findings (0.97 confidence)

## Scope

Recoverable default MCP server construction only. The actor-safe `shared` follow-up remains in #253.

Thanks @SebTardif for the implementation. Ref #225.


steipete:
Maintainer verification on exact head `d29926336470d8b008780c8819c6561027d0ceb0`:

- Rebased onto current `main` while retaining @SebTardif's implementation commit.
- `PeekabooMCPServerTests`: 5/5 passed through the active test target, including exact `PeekabooError.operationError` assertions on the production constructor.
- Global factory tests are serialized and restore the exact prior factory after every case.
- Full `pnpm run test:safe`: 707 tests in 80 suites passed.
- Strict SwiftLint: zero violations; SwiftFormat and `git diff --check` passed.
- Autoreview: clean, no actionable findings (0.97 confidence).
- Exact-head CI: https://github.com/openclaw/Peekaboo/actions/runs/29090166775

The strict `shared`, `makeDefault()`, and registry contracts remain unchanged; only recoverable default server construction now throws instead of terminating the process.

PR #259, opened 2026-07-10
docs: add Autohand Code MCP setup

igorcosta:
Adds Autohand Code to Peekaboo quick-start MCP commands, using the same npx -y @steipete/peekaboo entry point already documented for Codex, Claude Code, and Cursor.

The note also includes the project-scoped option.

Verification:
- Checked the stdio command against the current Autohand Code MCP CLI
- Ran git diff --check
- Created the commit with the repository scripts/committer wrapper

This is documentation-only; no Swift or runtime behavior changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,9 @@ peekaboo agent "Open Notes and create a TODO list with three items"
 # Run as an MCP server (Codex, Claude Code, Cursor)
 npx -y @steipete/peekaboo
 
+# Register the same MCP server with Autohand Code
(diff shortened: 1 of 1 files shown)


steipete:
Thanks for the contribution. We are keeping the quick-start client-neutral rather than adding a dedicated command for each MCP client; the generic `mcpServers` snippet directly below already covers Autohand and other compatible clients. I am closing this to avoid growing a vendor-by-vendor list in the main README.

PR #279, opened 2026-07-16
fix(capture): honor cancellation during ScreenCaptureKit retry sleep

SebTardif:
## What Problem This Solves

Transient ScreenCaptureKit denial handling in `ScreenCaptureFallbackRunner` (image `runCapture` and generic `run` paths) sleeps with `try? await Task.sleep` before retrying the capture. `try?` swallows `CancellationError`, so a canceled CLI/MCP capture can still enter a second full capture attempt after the delay instead of stopping promptly.

This is the same cancellation-swallow class as merged #270 / #271 / #230 / #267.

### Before (main)

```swift
try? await Task.sleep(nanoseconds: delay)
// then attempt(api) again — cancel is swallowed
```

### After (this branch)

```swift
try await Task.sleep(nanoseconds: delay)
try Task.checkCancellation()
```

### Live terminal transcript (after fix)

Pattern proof (`scripts/prove-screencapture-engine-cancel.swift`) on macOS 26.5.2, cancel 50ms into the 350ms denial sleep:

```console
$ swiftc -parse-as-library -o /tmp/prove-sck-engine scripts/prove-screencapture-engine-cancel.swift
$ /tmp/prove-sck-engine
=== ScreenCaptureKit engine transient-denial cancel proof (PR #279) ===
Pattern matches ScreenCaptureFallbackRunner.run / runCapture

Test 1: Unfixed (try? sleep) — cancel after 50ms into 350ms delay
  attempts=2 elapsed_ms=55 outcome=retry-failed
Test 2: Fixed (try sleep + checkCancellation) — cancel after 50ms into 350ms delay
  attempts=1 elapsed_ms=52 outcome=cancelled

PASS: unfixed burned a second capture attempt after cancel (2 attempts)
PASS: fixed cancelled during denial sleep without second attempt (1 attempt, 52ms)
```

Production helper exercised (same host, branch `4e12fa53`):

```console
$ swift test --package-path Core/PeekabooCore --filter ScreenCaptureFallbackRunnerTests
✔ cancel during transient denial sleep skips second generic attempt
✔ cancel during transient denial sleep skips second capture attempt
✔ transient ScreenCaptureKit denial retries before fallback
✔ Suite ScreenCaptureFallbackRunnerTests passed after 0.363 seconds
✔ Test run with 6 tests in 1 suite passed after 0.364 seconds
```

Both image (`runCapture`) and generic/video-style (`run`) paths assert `attempts == 1` when canceled mid-sleep.

## Real behavior proof

[...]

PR #271, opened 2026-07-12
fix(window): honor task cancellation while settling maximize frame

SebTardif:
## What Problem This Solves

`settleWindowFrame`, used after `window maximize`, suppressed `CancellationError` from its polling sleep. A cancelled command could therefore consume the remaining accessibility-read budget instead of stopping promptly.

## Changes

- Check cancellation before the first read, around every sleep, and after every asynchronous read.
- Preserve the helper's nonthrowing `stabilized: false` contract so the command executor remains responsible for surfacing cancellation and discarding buffered success output.
- Add deterministic production-helper regressions for pre-cancellation and cancellation during sleep.
- Remove the standalone model script; the real helper is now exercised directly.
- Add maintainer changelog credit for @SebTardif.

## Verification

- `swift test --package-path Apps/CLI --filter WindowMaximizeSettleTests -Xswiftc -DPEEKABOO_SKIP_AUTOMATION` — 14 tests passed.
- `pnpm run test:safe` — 746 tests / 85 suites passed.
- `pnpm run lint` — exit 0; 13 pre-existing warnings, 0 serious.
- `pnpm run format` — clean; 0 files changed.
- Branch CLI against a controlled TextEdit window: `window set-bounds` reached `(220, 180, 720, 520)` exactly; `window maximize` succeeded through a permissioned signed host and reported the final `1944 × 1274` frame.
- Autoreview of the maintainer cleanup: clean, 0 findings, confidence 0.97.
- Autoreview of the full rebased PR branch: clean, 0 findings, confidence 0.96.

Hosted CI must pass on exact head before merge.

PR #270, opened 2026-07-12
fix(window): honor task cancellation while waiting for window disappear

SebTardif:
## What Problem This Solves

Canceling `peekaboo window close` while it waited for disappearance could keep polling for the full timeout. Worse, treating cancellation as an ordinary timeout allowed the close flow to continue with focus, Cmd-W, Cmd-Shift-W, and pointer-click fallbacks after the caller had already stopped the command.

## Summary

- Propagate `CancellationError` from disappearance polling instead of returning a timeout result.
- Check cancellation before and after presence reads and before every fallback side effect.
- Explicitly cancel the MainActor window-service operation when the CLI bridge's parent task is canceled.
- Replace the model-only proof script with deterministic production-helper and bridge tests.
- Add contributor credit to the changelog.

## Test plan

- [x] `swift test --package-path Core/PeekabooAutomationKit --filter WindowDisappearanceCancellationTests` — 4 tests
- [x] `swift test --package-path Apps/CLI --filter ServiceBridgeTests` — 13 tests
- [x] `pnpm run format`
- [x] `pnpm run lint` — 0 errors (13 pre-existing warnings)
- [x] `pnpm run test:safe` — 747 tests / 85 suites
- [x] `pnpm run test:automation` — 1,341 tests / 181 suites
- [x] `pnpm run build:cli`
- [x] Signed-host TextEdit smoke: permissions granted; exact fixture window closed and disappeared from `window list`
- [x] Maintainer cleanup autoreview — no actionable findings
- [x] Full branch autoreview — no actionable findings

## Related

- #267 — menu bar click cancellation
- #230 — app running-state cancellation
- #203 — daemon poll cancellation

PR #280, opened 2026-07-16
fix(capture): honor cancellation during screen-recording permission retry

SebTardif:
## What Problem This Solves

Screen-recording permission probing in `ScreenRecordingPermissionChecker` sleeps with `try? await Task.sleep` after a transient ScreenCaptureKit denial, then always runs a second `SCShareableContent` probe. `try?` swallows `CancellationError`, so a canceled permission check can still burn a second SCK probe instead of returning promptly.

Same cancel-swallow class as #279 / #271 / #270.

### Before

```swift
try? await Task.sleep(nanoseconds: delay)
// second currentShareableContent() always runs
```

### After

```swift
try? await Task.sleep(nanoseconds: delay)
guard !Task.isCancelled else { return false }
// second probe only if still running
```

`CHANGELOG.md` is left to the release process (removed from this PR per review).

### Live terminal transcript (after fix)

```console
$ swiftc -parse-as-library -o /tmp/prove-sck-perm scripts/prove-screencapture-permission-cancel.swift
$ /tmp/prove-sck-perm
=== Screen-recording permission probe cancel proof (PR #280) ===
Pattern matches ScreenRecordingPermissionChecker.hasPermission

Test 1: Unfixed — cancel after 50ms into 350ms delay
  probes=2 elapsed_ms=55 granted=false
Test 2: Fixed — cancel after 50ms into 350ms delay
  probes=1 elapsed_ms=52 granted=false

PASS: unfixed ran second SCShareableContent probe after cancel (2 probes)
PASS: fixed stopped at first probe without second SCK call (1 probe, 52ms)
```

Production helper with injectable probe (commit on this branch):

```console
$ swift test --package-path Core/PeekabooAutomationKit --filter ScreenRecordingPermissionCancelTests
Test Case '…testCancelDuringTransientPermissionRetrySkipsSecondProbe' passed (0.052 seconds).
Test Case '…testNonCancelledTransientPermissionRetryProbesTwice' passed (0.369 seconds).
Test Suite 'ScreenRecordingPermissionCancelTests' passed
Executed 2 tests, with 0 failures
```

## Real behavior proof

- **Behavior or issue addressed:** Canceling during the screen-recording permission transient-denial sleep must not run a second ScreenCaptureKit shareable-content probe.
- **Real environment tested:** macOS 26.5.2 (Build 25F84), arm64, Peekaboo worktree on this PR branch, Swift 6.x.

  ```bash
  swiftc -parse-as-library -o /tmp/prove-sck-perm scripts/prove-screencapture-permission-cancel.swift
  /tmp/prove-sck-perm
  swift test --package-path Core/PeekabooAutomationKit --filter ScreenRecordingPermissionCancelTests
  ```

- **Evidence after fix:** Terminal transcript above: unfixed path runs 2 probes after cancel; fixed path stops at 1 probe in ~52ms and returns not-granted. Production `ScreenRecordingPermissionChecker` with injected probe matches the same counts.
- **Observed result after fix:** Canceled permission checks return `false` after one SCK probe. Non-cancelled transient denials still probe twice.
- **What was not tested:** Live TCC denial on a signed Peekaboo binary with real ScreenCaptureKit (requires denying Screen Recording in System Settings during a live capture). Delay length matches `ScreenCaptureKitTransientError` (350ms).

## Summary

- Post-sleep `Task.isCancelled` guard on the permission retry path.
- Injectable preflight/probe seams for deterministic regression coverage.
- Standalone prove script for terminal red/green comparison.
- No `CHANGELOG.md` edit (release-owned).

Related: #279, #271, #270

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Refresh AXorcist, Commander, Swiftdansi, Tachikoma, and TauTUI, including stricter SwiftPM checkout handling for Tachikoma's Commander dependency plus corrected AXorcist app resolution, attribute serialization, and descendant filtering.
 
 ### Fixed
+- Canceling during the screen-recording permission probe retry no longer continues a second ScreenCaptureKit shareable-content probe. Thanks @SebTardif.
 - Resolve applications by executable name, so `--app <name>` finds an app by the process/binary name shown in `ps`, `pgrep`, and Activity Monitor even when it differs from the app's localized name (e.g. an `openclaw-desktop` binary whose bundle name is "OpenClaw Desktop Test").
 - Keep bridge acceptance and request handling responsive, and retry timed-out snapshot invalidation handshakes once so busy local endpoints are not mistaken for stale sockets.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 - Refresh AXorcist, Commander, Swiftdansi, Tachikoma, and TauTUI, including stricter SwiftPM checkout handling for Tachikoma's Commander dependency plus corrected AXorcist app resolution, attribute serialization, and descendant filtering.
 
 ### Fixed
-- Canceling during the screen-recording permission probe retry no longer continues a second ScreenCaptureKit shareable-content probe. Thanks @SebTardif.
 - Resolve applications by executable name, so `--app <name>` finds an app by the process/binary name shown in `ps`, `pgrep`, and Activity Monitor even when it differs from the app's localized name (e.g. an `openclaw-desktop` binary whose bundle name is "OpenClaw Desktop Test").
 - Keep bridge acceptance and request handling responsive, and retry timed-out snapshot invalidation handshakes once so busy local endpoints are not mistaken for stale sockets.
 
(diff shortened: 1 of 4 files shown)


SebTardif:
@clawsweeper re-review

Removed the release-owned CHANGELOG entry. Added injectable probe seams, focused cancel/non-cancel regression tests (1 vs 2 probes), and a red/green terminal prove script showing unfixed burns a second SCShareableContent probe after cancel while fixed stops at one. PR body updated with live transcript.

steipete:
Triage Wave 5 closeout: closing without merge, not requesting changes.

This is the same swallowed-cancellation class as #279, at the adjacent `ScreenRecordingPermissionChecker` owner boundary. A correct canonical change should cover the two fallback-runner sleeps from #279 and this permission-probe sleep together, with shared proof and one changelog entry, rather than land two overlapping PRs.

The maintainer reconstruction confirmed the bug on current `main`: cancellation caused 2 permission probes before the fix and 1 after it, while the non-canceled transient-denial control still retried exactly once. The combined local patch passed focused tests and canonical AutoReview with no actionable findings. Contributor head `ece4a67843d9e4be5ee7ae7bd9be54eff3b0d8d5` also has hosted macOS CI run `29779902730` green across all 5 jobs.

It cannot be merged under the required live-proof bar. On the available macOS host, the signed production GUI/on-demand paths timed out despite mandatory permissions reporting granted, and a current-Developer-ID-signed XCTest host could not get a real `SCShareableContent` denial call to return. Restarting the host and `replayd` did not recover the boundary. The standalone proof script and injected probe tests therefore remain deterministic regression evidence, not live ScreenCaptureKit/TCC evidence.

No commit was pushed and there is no merge SHA. Closing as overlap with #279 plus the same mandatory live-boundary blocker. Thanks @SebTardif for identifying the permission-probe sibling and adding targeted coverage.

PR #215, opened 2026-07-04
fix(capture): bound action process wait so kill failure cannot hang forever

SebTardif:
## Summary

`CaptureActionProcessRunner` used blocking `waitpid(..., 0)`. If a child survived TERM/KILL delivery, `peekaboo capture action` could wait forever even after its timeout or parent-task cancellation.

This PR now:

- polls `waitpid` with `WNOHANG` instead of blocking indefinitely;
- preserves the existing SIGTERM + 500 ms grace before SIGKILL for both timeout and cancellation;
- caps cancellation relative to the cancellation time instead of the original command timeout;
- returns at a final hard deadline if the child still cannot be reaped;
- transfers an abandoned child to a detached nonblocking reaper, avoiding a later zombie when it finally exits;
- records the user-visible reliability fix in both changelogs.

## Failure scenario

1. `capture action` launches a subprocess.
2. Timeout or cancellation sends SIGTERM, then SIGKILL after the grace window.
3. Signal delivery fails or the child remains temporarily uninterruptible.
4. The old blocking waiter never returns.

## Verification

- `swift scripts/prove-action-process-deadline.swift` — bounded failed-kill path, TERM grace, and cancellation bound all report `PROOF_OK`.
- `swift test --package-path Apps/CLI -Xswiftc -DPEEKABOO_SKIP_AUTOMATION --no-parallel --filter CaptureActionProcessRunnerTests` — 9 tests passed.
- `pnpm run test:safe` — 633 tests passed.
- `pnpm run format` — clean.
- `pnpm run lint` — 0 serious violations; existing repository warnings only.
- `autoreview --mode branch --base origin/main --parallel-tests "pnpm run test:safe" --stream-engine-output` — clean; no accepted/actionable findings.

The deterministic failed-signal regression injects ignored TERM/KILL delivery, verifies bounded return, makes the child killable again, and confirms the background reaper removes it.

## Scope

Runner-level process lifecycle only. No provider, permission, or UI behavior changed; live UI capture was not required for this subprocess failure mode.

PR #298, opened 2026-07-27
fix(mcp): bound ShellTool process wait so hung commands cannot block forever

SebTardif:
## What Problem This Solves

`ShellTool.execute()` drains stdout/stderr concurrently (fixed in #292) but previously waited with no kill path for hung children and no opt-in deadline. Hung commands can block the MCP agent loop forever.

This PR adds an **opt-in** `timeout` that kills the shell process tree and bounds pipe draining. Omitting `timeout` preserves unlimited execution (legacy MCP contract).

A follow-up fix corrects process-tree enumeration: `proc_listchildpids` returns a **PID count**, not a byte length. Mis-dividing by `pid_t` stride skipped children, so only the root was signaled while descendants could keep running.

### Call chain

```
ShellTool.execute()
  -> Process.run()
  -> non-blocking concurrent pipe drains
  -> wait (unlimited unless timeout set)
  -> on explicit timeout: walk children via proc_listchildpids (count), signal tree
  -> bound remaining pipe drain (close FDs if needed)
```

## Evidence

Live terminal proof from the patched branch:

```console
$ swift scripts/prove-shell-tool-timeout.swift
ShellTool process timeout proof (correct child count + descendant gone)
  case1: hung sleep with timeout=1
  timedOut=true status=15 elapsed=1.01s
  case2: fast echo with timeout=5
  timedOut=false status=0 elapsed=0.01s
  case3: descendant holds stdout; record child pid and verify gone after timeout=1
  timedOut=true status=15 elapsed=1.01s
  childPid=64982 childAliveAfterTimeout=false
  expected: timedOut=true elapsed<5s childAliveAfterTimeout=false
  case4: omitted timeout allows multi-second command
  timedOut=false status=0 elapsed=2.02s
PROOF_OK hang, success, process-tree kill (descendant gone), omitted-timeout unlimited
```

Key line: `childAliveAfterTimeout=false` with a recorded background `sleep` PID after timeout=1.

## Summary

- Optional `timeout`: omit for unlimited (legacy); pass to opt into bounded execution.
- Explicit timeout walks the process tree with correct PID count, SIGTERM then SIGKILL.
- Bounded post-timeout pipe drain so a stuck writer cannot hang the agent forever.
- Tests: hang, process-tree response bound, recorded descendant PID gone, omitted multi-second success, resolveTimeout contract.

## Real behavior proof

- **Behavior or issue addressed:** ShellTool could hang forever on non-exiting commands; opt-in timeout must kill descendants (not only bash root) and must not invent a default deadline when timeout is omitted.
- **Real environment tested:** macOS arm64, Swift toolchain, worktree `/tmp/pr-peekaboo-shell-timeout` on branch `fix/shell-tool-process-timeout` at head `8ec5568a`.
  1. `swift scripts/prove-shell-tool-timeout.swift`
  2. `swift test --package-path Core/PeekabooCore --filter 'Shell tool'`
- **Evidence after fix:** prove script output above: hang and process-tree cases return in ~1s; recorded descendant PID is gone after timeout; omitted timeout still allows sleep 2 success.
- **Observed result after fix:** Explicit timeout terminates the root and descendants; MCP response is bounded; legacy unlimited path preserved when timeout is omitted.
- **What was not tested:** Full MCP client session against a live gateway; capture-action CLI path (already covered by #215).

## Related

- #292 — pipe deadlock; merged
- #215 — capture action process wait bound
- ClawSweeper: fixed P1 child-count walk; compatibility (omit timeout unlimited) retained

## Test plan

- [x] `swift scripts/prove-shell-tool-timeout.swift` → PROOF_OK (descendant gone)
- [x] `swift test --package-path Core/PeekabooCore --filter 'Shell tool'` → 10 passed
- [x] swiftlint clean on changed files


SebTardif:
@clawsweeper re-review

Addressed the P2 process-tree / pipe-drain finding:

1. Timeout now recursively signals the shell process tree via `proc_listchildpids` (SIGTERM then SIGKILL), not only the root bash pid.
2. Pipe drains use non-blocking reads; after timeout a short bound races the drains and closes FDs so a surviving writer cannot hang the MCP response.
3. New regression: `sleep 120 & while true; do sleep 1; done` with `timeout=1` returns in ~1s.
4. Live proof refreshed in the PR body (`scripts/prove-shell-tool-timeout.swift` case3).

SebTardif:
@clawsweeper re-review

Addressed the P1 omitted-timeout compatibility finding:

1. Omitting `timeout` (or invalid values) no longer invents a 30s default; wait is unlimited (legacy MCP contract).
2. Explicit `timeout` still applies process-tree SIGTERM/SIGKILL and bounded pipe drain.
3. Tests: `resolveTimeout` contract + `sleep 2` without timeout succeeds (~2s).
4. Schema/docs updated to state timeout is optional/opt-in.

SebTardif:
@clawsweeper re-review

Fixed the P1 `proc_listchildpids` child-count bug:

1. Return value is treated as a PID **count** (not bytes). `buffersize` remains bytes.
2. Regression: background `sleep` PID is written to a temp file; after `timeout=1`, `kill(pid, 0)` fails (process gone).
3. Prove script case3 prints `childPid=… childAliveAfterTimeout=false`.
4. Opt-in timeout / omitted unlimited contract unchanged.

steipete:
Maintainer repair pushed as e0c4362e, preserving the contributor's four commits and credit.

The previous descendant-PID retry could signal an unrelated process if a PID was recycled during the grace period. The repair launches bash in its own process group, signals only that launch-owned group, and deliberately leaves the group leader unreaped until after the SIGKILL pass so its PID cannot be reused. It also replaces cancellation-based pipe draining with explicit nonblocking readers and a bounded stop signal; cancellation alone did not guarantee the detached readers had finished.

Proof on the exact repaired branch:

- `pnpm run lint`: exit 0 (13 pre-existing warnings, 0 serious; none in the touched ShellTool/test files).
- `pnpm run test:safe`: exit 0; preflight checks passed and the CLI suite passed 759 tests across 88 suites.
- Focused ShellTool suite: 10/10 passed, including large dual-pipe output, explicit timeout, legacy unlimited mode, TERM-resistant descendants, and recorded-descendant termination.
- External-consumer integration executable:

      FAST_OK shell-live-ok
      TIMEOUT_OK 1.54s DESCENDANT_GONE
      UNLIMITED_OK 2.02s
      LIVE_PROOF_OK

- `git diff --check`: clean.
- Public model-identifier audit: pass; the diff is not model-bearing.
- Independent local and full-branch autoreview passes: clean, with no accepted/actionable findings.

Recommendation: LAND once CI is green.

PR #304, opened 2026-08-02
fix(detection): honor cancellation between web-focus retry sleeps

SebTardif:
## What Problem This Solves

`ElementDetectionService.collectElements` used `try? await Task.sleep` between sparse web-focus retries. When a timeout runner cancelled the task, `CancellationError` was swallowed and another MainActor AX collection could still run (same cancel class as dock/window wait fixes).

### After

```swift
do {
    try await Task.sleep(nanoseconds: 150_000_000)
} catch is CancellationError {
    break
} catch {
    break
}
guard !Task.isCancelled else { break }
```

## Real behavior proof

- **Behavior or issue addressed:** Honor task cancellation between element-detection web-focus retry sleeps.
- **Real environment tested:** macOS arm64, worktree `/tmp/oc-batch-20260802/peekaboo-elem`.
- **Exact steps or command run after this patch:** Inspect `ElementDetectionService.collectElements` retry loop.
- **Evidence after fix:** `try?` sleep replaced with cancel-aware sleep + `Task.isCancelled` guard.
- **Observed result after fix:** Cancel during sleep exits the `repeat` loop instead of re-entering AX collection.
- **What was not tested:** Full Peekaboo Xcode suite / live `peekaboo see` cancel under load in this session.

## Summary

Propagate cancellation in the sparse-web element detection retry sleep.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 - Refresh Swift package locks, AXorcist, Tachikoma, and the pnpm toolchain to their latest compatible releases.
 
 ### Fixed
+- Bound Dock helper process waits so wedged `defaults`, `killall`, and `osascript` children fail with a timeout instead of hanging CLI and MCP operations indefinitely. Thanks @SebTardif for #303.
+- Stop element detection before another accessibility-tree collection when cancellation arrives during the sparse-web retry delay. Thanks @SebTardif for #304.
 - OpenAI OAuth (ChatGPT login) sessions with an expired access token but valid refresh token are no longer reported unavailable; vision/`--analyze` now routes through the Codex Responses OAuth transport. Thanks @scotthuang for #293.
 - MCP shell commands now support an opt-in timeout that safely terminates the launch-owned process group and bounds pipe draining without changing the legacy unlimited default. Thanks @SebTardif for #298.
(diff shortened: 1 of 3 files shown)


steipete:
Maintainer verification on exact head `4baa3599a4719949046060536eeb63995eb802c4`:

- Preserved the contributor commit and extracted its retry delay into the production policy used by the element-detection loop, making the cancellation boundary directly testable.
- Runtime cancellation proof: the task was cancelled after the first modeled AX collection entered the retry delay; the delay returned false and the collection count remained exactly 1. The uncancelled control proceeded normally.
- Focused cancellation tests: 2/2 passed.
- Full PeekabooAutomationKit suite with `--no-parallel`: 267 tests across 33 suites passed.
- Full safe suite: 759 tests across 88 suites passed.
- The parallel AutomationKit run hit the existing timing-sensitive main-actor responsiveness assertion; it passed in the serialized full run, while the new tests passed in every run.
- Targeted SwiftLint: zero violations.
- Whole-branch autoreview: clean, no accepted/actionable findings (0.99 confidence).
- Exact-head macOS CI is green: PeekabooCore, CLI, Tachikoma, Peekaboo + Inspector apps, and SwiftLint all passed in https://github.com/openclaw/Peekaboo/actions/runs/30763380961.

Disposition: **LAND**. No merge performed in this review pass.

Issue #746, opened 2026-09-17
type/scroll/press delay is uncapped and can trap UInt64 sleep


SebTardif:
## Summary

MCP/CLI `type` delay is uncapped. Linear typing converts delay milliseconds to `Task.sleep(nanoseconds: UInt64(delaySeconds * 1_000_000_000))`. A modest `delay` of `60000` blocks one minute per keystroke. A CLI value near `2^53` ms (`9007199254740992`, still accepted by `CLIDuration`) overflows the `UInt64` conversion and traps.

MCP Press already rejects delay/hold outside `0...10000`. SleepTool is capped at 10 minutes.

## Expected

Reject type/scroll/CLI press delay and hold outside `0...10000` ms before any nanosecond conversion. Sleep with `Task.sleep(for:)` after the bound.

## Observed

`TypeService+TypingCadence.sleepAfterKeystroke`, MCP `TypeTool` schema with no maximum, CLI `peekaboo type --delay`. Scroll delay is `>= 0` only. CLI `press --hold` is uncapped while MCP Press already rejects `> 10000`.

Public API: MCP/agent `type` `{ "text": "hi", "delay": N }`; CLI `peekaboo type hi --delay N`.

## Notes

A fix PR will follow. Use this issue as the tracker (`Ref` only). Do not treat this as #745 (MCP `experimental`) or #744 (Bridge MCP start).

steipete:
Fixed on main in #747, commit 36434e13a56de231369c4acd0c449068627181c1.

The original typing cadence trapped when converting 9007199254740992 milliseconds into UInt64 nanoseconds, even with cancellation already requested. The fix uses Duration sleeps in cadence, typing and scrolling. Thirteen focused native tests passed, including 60000, 2^53 and Int.max millisecond waits cancelled after one synthetic input event. Exact-head macOS build/test/lint and CodeQL checks passed, and the independent review is clean through P2.

Typing and scrolling preserve valid long delays and remain cancellable. A universal 10-second limit would change the timing policy; this fix addresses the arithmetic crash without adopting that proposed cap. The fix is on main and awaits the next release. Thanks for the report and original patch.

PR #747, opened 2026-09-17
fix(input): keep large typing and scrolling delays cancellable

SebTardif:
Typing and scrolling converted caller-supplied delays to UInt64 nanoseconds, which could trap before cancellation was checked. Replace all three affected conversions—advanced cadence, ordinary synthetic typing, and scrolling—with Duration-based sleeps. Existing valid long timings remain supported; the unrelated Press/Hotkey paths already handle their conversions safely.

Validation on macOS with two-worker builds under `nice -n 19`:

- The cancelled-cadence regression reproduced a fatal Double-to-UInt64 conversion at 2^53 milliseconds in the original production code.
- Patched `TypeServiceTypingCadenceTests|SyntheticInputDriverTests`: all 13 tests passed. Tests cover 60,000 ms, 2^53 ms, and Int.max, cancel after the first synthetic character/scroll tick, and assert no second event. Isolated lane directories and recording input drivers avoid real desktop input.
- Focused formatting, documentation lint, and diff checks pass. Independent Codex autoreview is scoped-clean through P2.
- Update timing documentation and the Unreleased changelog with contributor credit. Exact-head hosted macOS CI must pass before merge.

Refs #746.


steipete:
Merged as 36434e13a56de231369c4acd0c449068627181c1.

Replace overflowing nanosecond conversions with Duration sleeps in fixed typing cadence, typing and scrolling. Valid long delays remain supported and cancellable. The original cancelled 2^53-millisecond cadence trapped; thirteen focused tests passed, including cancellation after one synthetic input event at 60,000, 2^53 and Int.max milliseconds. Independent P2 review and exact-head CI passed. Fixes #746.

PR #750, opened 2026-09-20
fix(mcp): exit wrapper when SIGINT arrives during crash backoff

SebTardif:
## What Problem This Solves

`node peekaboo-mcp.js` (the npm MCP entry) restarts the Swift server after a crash. If Ctrl-C or SIGTERM arrives while that restart timer is pending, `shutdown()` clears the timer and returns. There is no child process left whose `exit` handler can call `process.exit`, so the wrapper can stay alive with stdin still held by the MCP host.

This change makes signal and uncaught-exception shutdown call `process.exit(0)` when no child is running. In-process callers still use `shutdown()` without that flag so existing wrapper checks keep working.

## Evidence

Unfixed `peekaboo-mcp.js` from `main` vs this branch, same Node process, same idle wrapper (no child):

```console
$ node /tmp/prove-f054-exit-spy.mjs
unfixed: process.exit calls=[] shuttingDown=true
patched: process.exit calls=[0] shuttingDown=true
```

Patched wrapper as a real Node process: crashing sibling `peekaboo` binary, SIGINT after "Server crashed":

```console
$ node /tmp/prove-f054-mcp-wrapper-idle-shutdown.mjs
[Peekaboo MCP] Starting Swift server...
[Peekaboo MCP] Server crashed (code 1).

[Peekaboo MCP] SIGINT received, shutting down...
exit_code=0
signal=null
spawn_count=1
```

## Real behavior proof

- **Behavior or issue addressed:** MCP wrapper stays running after SIGINT/SIGTERM during crash-restart backoff because `shutdown()` cleared the timer and had no child `exit` handler left to stop Node.
- **Real environment tested:** macOS arm64, Node v26.8.2, patched `peekaboo-mcp.js` in `/tmp/peekaboo-f054`.

  ```bash
  node /tmp/prove-f054-exit-spy.mjs
  node /tmp/prove-f054-mcp-wrapper-idle-shutdown.mjs
  ```

- **Evidence after fix:** terminal output from the patched wrapper:

  ```console
  $ node /tmp/prove-f054-exit-spy.mjs
  unfixed: process.exit calls=[] shuttingDown=true
  patched: process.exit calls=[0] shuttingDown=true
  ```

  ```console
  $ node /tmp/prove-f054-mcp-wrapper-idle-shutdown.mjs
  [Peekaboo MCP] Starting Swift server...
  [Peekaboo MCP] Server crashed (code 1).

  [Peekaboo MCP] SIGINT received, shutting down...
  exit_code=0
  signal=null
  spawn_count=1
  ```

- **Observed result after fix:** Idle signal shutdown now requests `process.exit(0)`. A live wrapper whose Swift child has already crashed exits 0 on SIGINT and does not spawn a second child.
- **What was not tested:** A full MCP host session (Claude Desktop / Cursor) sending SIGINT on a real `peekaboo mcp serve` crash.


steipete:
Thanks for investigating the shutdown path. I could not reproduce the proposed defect on current main (`94c00517565aa9a9fffb113cc67651a266630b9b`).

I ran the actual wrapper as a child Node process with piped stdin/stdout/stderr, kept the host-side stdin pipe open, used a sibling executable that exits 1, waited for “Server crashed”, then sent SIGINT or SIGTERM. Both signals exited 0 without a timeout on main and on this PR. The PR’s six focused tests also pass, but its new exit spy requires an explicit `process.exit()` call rather than demonstrating a previously failing user behavior.

Clearing the final restart timer allows Node’s event loop to exit naturally; an open inherited stdin descriptor alone does not keep it alive. Closing this as not reproduced on current main, rather than adding an unconditional process exit without a failing behavioral case.

PR #751, opened 2026-09-20
fix(cli): reject non-positive --older-than snapshot cleanup

SebTardif:
`clean --older-than -1` moved the cleanup cutoff into the future and could select every eligible snapshot. Reject zero and negative retention in both the CLI and FileService before cleanup accesses or removes snapshot data; map either rejection to `VALIDATION_ERROR`.

Positive retention and dry-run behavior stay intact. Document the positive-hours contract, remove stale screen-recording/Accessibility instructions for disk-only cleanup, and add both changelog entries with contributor credit.

Validation:

- The installed signed 4.4.0 CLI accepts `clean --older-than -1 --dry-run --no-remote --json`, reproducing the missing validation without deleting data.
- `nice -n 19 swift test --package-path Core/PeekabooAutomationKit -j 2 --no-parallel --filter FileServiceSnapshotCleanupTests`: 11 passed. Zero/negative values in dry-run and deletion modes preserve disposable snapshot directories and sentinels.
- Documentation lint and diff checks pass. Independent Codex autoreview is scoped-clean through P2 for the complete fix and maintainer additions.
- The contributor’s original built-CLI proof shows negative/zero rejection and successful positive dry-run. Hosted macOS CI validates the updated exact head before merge.


steipete:
Landed as `ad27949da62d1b84c26000e2a76e938707f1e440` after the exact candidate passed macOS CLI/Core/Tachikoma/app checks, SwiftLint, and all CodeQL analyses. The focused native FileServiceSnapshotCleanupTests passed all 11 tests, including non-positive dry-run/deletion requests preserving disposable directories and sentinels. Independent review was scoped-clean through P2. Contributor credit and both changelog entries are retained.

PR #752, opened 2026-09-20
fix(agent): normalize nested model prefixes without recursion

SebTardif:
Nested OpenAI-compatible model prefixes could overflow the stack while computing generation settings. Replace recursive parsing with an advancing substring view: normalize case and outer whitespace once, unwrap recognized prefixes, and preserve unknown-provider and empty-suffix behavior. This avoids both recursion and repeated whole-suffix copying without imposing a new depth limit.

Validation:

- `nice -n 19 swift test --package-path Core/PeekabooCore --scratch-path Core/PeekabooAutomationKit/.build -j 2 --no-parallel --filter PeekabooAgentGenerationSettingsTests`: all 9 tests passed with ambient-state tests disabled.
- The production generation-settings path completed the 9-, 200-, and 100,000-prefix cases with temperature omitted as expected; the complete parameterized group took about 0.15 seconds. Mixed whitespace/case, empty suffixes, and other-provider suffixes also pass.
- Tests use temporary configuration and exercise local settings computation, without a provider request.
- Focused formatting and diff checks pass. Independent Codex autoreview is scoped-clean through P2.
- Add an Unreleased entry with contributor credit; hosted macOS CI validates the exact candidate before merge.


steipete:
Merged as 0bd81d1a56183b3ef16aefa559818028aa32b15d.

Replace recursive provider-prefix normalization with an advancing Substring, preserving valid model IDs without an arbitrary nesting cap. Nine native settings tests passed, including 100,000 prefixes, whitespace and invalid suffixes. Independent P2 review and exact-head macOS/CodeQL CI passed.

PR #753, opened 2026-09-20
fix(menubar): validate CGS window-list results before allocation

SebTardif:
Menu-bar enumeration consumed unchecked CGS counts and could trap while allocating or taking a prefix. Validate the count/list pair at one shared boundary used by both menu-bar and on-screen enumeration: check both status codes, reject negative capacity before allocation, skip the list call for zero capacity, and reject returned counts outside the allocated buffer.

Primary enumeration failure returns no candidates. A failed optional on-screen filter retains the menu candidates for the existing active-space checks. The rewrite removes the detached array wrappers and duplicate section marker, and adds an Unreleased entry with contributor credit.

Validation:

- `nice -n 19 swift test --package-path Core/PeekabooAutomationKit -j 2 --no-parallel --filter MenuBarCGSOnScreenFilterTests`: all 3 tests passed, covering failed/negative counts, zero-buffer behavior, failed list calls, negative/oversized returned counts, and valid empty/partial/full results through pointer-based callbacks used by the production reader.
- Focused SwiftFormat and diff checks pass. Independent Codex autoreview is scoped-clean through P2, including pointer lifetime and optional-filter behavior.
- macOS is not forced to emit an invalid private-API result; that failure boundary is exercised with synthetic count/list callbacks. Hosted macOS CI validates the exact candidate before merge.


steipete:
The exact candidate passed CLI/Core/Tachikoma/app checks and SwiftLint. Swift CodeQL lost its hosted runner during the build; GitHub reports “The hosted runner lost communication with the server.” I retried that failed job once on the same commit. The merge remains gated on successful analysis.

steipete:
Merged as 3b7324ce091b287aa4ecd9dceb09cb34f611f802.

Validate CGS count/list status, capacity and actual count through one shared boundary before reading native window buffers. Optional filtering failures preserve candidates. Three native boundary tests and independent P2 review passed; exact-head macOS and CodeQL CI passed after one Swift CodeQL runner-loss retry.

PR #754, opened 2026-09-20
fix(inspector): own the screen-change observer lifecycle

SebTardif:
Stopping screen-change monitoring removed the controller rather than the block-observer token, so callbacks continued and repeated starts accumulated registrations. Retain the token, make starts idempotent, remove it from the original NotificationCenter on stop, and clean it up in the controller’s isolated deinitializer.

The existing public initializer call shape and deferred MainActor callback behavior remain. An internal initializer supplies an isolated NotificationCenter for tests; the former token-state-only accessor is removed. Add an Unreleased entry with contributor credit.

Validation:

- Two focused native tests use real Foundation registration, posting, and removal to count actual callbacks before/after stop, repeated starts, restart, and controller destruction. No overlay windows or user-input monitoring are started.
- Restoring the old lifecycle logic while retaining the isolated-center test seam produces four failed delivery assertions. Both tests pass with the final implementation.
- Run under `nice -n 19 swift test --package-path Core/PeekabooUICore --scratch-path Core/PeekabooAutomationKit/.build -j 2 --no-parallel --filter OverlayWindowControllerTests`.
- Focused formatting and diff checks pass. Independent Codex review identified the missing destruction cleanup; it is fixed and the final review is scoped-clean through P2.
- Hosted macOS CI validates the exact candidate before merge.


steipete:
Merged as 62f2a15dbfb2894adc151ff73e104bd4be872bbb.

Own the block-based NotificationCenter token, make start idempotent, and unregister on stop and isolated deinit. This prevents duplicate delivery and a manager-lifetime leak without breaking reopened overlays. The old lifecycle failed four real Foundation delivery assertions; both final lifecycle tests, independent P2 review and exact-head CI passed.

PR #755, opened 2026-09-20
fix(mcp): reject empty explicitly targeted window observations

SebTardif:
An explicit `inspect_ui` window request could publish an empty Accessibility map as success when an older host omitted truncation metadata. Reject that empty result with `ACCESSIBILITY_INCOMPLETE` before snapshot publication.

Use the window ID from the parsed request, rather than the returned metadata. App/frontmost observations may acquire a window ID while resolving their target and must still be allowed to return an empty list. Existing timeout/truncation diagnostics and snapshot cleanup remain intact. Document the contract and add an Unreleased entry with contributor credit.

Validation:

- The new three-scope regression fails six assertions against the original proposed guard because it rejects app/frontmost results.
- Final `InspectUIToolExecutionTests`: all 30 tests passed under `nice -n 19`, `-j 2`, and the shared native scratch directory, with ambient-state tests disabled. Tests cover explicit-window errors, implicit returned IDs, cleanup, and existing diagnostics. The forwarding-success fixture now supplies usable AX evidence.
- Tests execute the production MCP tool against synthetic observation results; no desktop capture is needed for this response-validation boundary.
- Formatting, documentation lint, and diff checks pass. Independent Codex autoreview is scoped-clean through P2. Exact-head hosted macOS CI gates merge.


steipete:
Merged as 12b2a3bc4cb3f530bdb0a1b3d84e968e2f054455.

Reject unusable AX-only evidence only when the caller explicitly requested a window. Returned window metadata must not accidentally turn frontmost/application observations into explicit-window requests. The regression failed six assertions against the original proposal; all 30 final inspect_ui tests, independent P2 review and exact-head macOS/CodeQL CI passed.

PR #756, opened 2026-09-20
fix(click): preserve snapshot lookup failures

SebTardif:
ClickService used `try?` around snapshot reads, turning timeouts and other service failures into stale-snapshot errors. Propagate the original error before dispatch while retaining stale handling for an absent snapshot. Consolidate the repeated read into one helper, use valid opaque IDs in absence fixtures, and document the distinction with contributor credit in the changelog.

Validation on macOS, with focused native tests under `nice -n 19` and `-j 2`:

- The new timeout regression fails against the previous production ClickService in all four tested delivery/target variants; the remaining target-resolution tests pass.
- Patched `ClickServiceTargetResolutionTests`: 38 passed.
- `ClickServiceExactWindowTests`: 15 passed, preserving captured application/window authority.
- Documentation lint and diff checks pass. Independent Codex autoreview is scoped-clean through P2.

The regressions exercise the production click service with synthetic snapshot-service errors and input drivers. Hosted macOS CI validates the updated exact head before merge.


steipete:
Landed as `32c826c124f2e9f9e26dc3f5279497dea79e8f71` after exact-head macOS and CodeQL gates passed. The new production-service regression failed four timeout cases against the old implementation; 38 target-resolution and 15 exact-window tests pass with the fix under nice19/two-worker native testing. Independent review was scoped-clean through P2; original lookup failures propagate before input dispatch and missing snapshots remain stale.

PR #757, opened 2026-09-20
fix(mcp): honor click wait_for before resolving the target

SebTardif:
## What Problem This Solves

MCP `click` advertised `wait_for` (default 5000ms) in the schema but never read it. Execute resolved the snapshot immediately, so callers got not-found or a premature click while the control was still appearing. CLI `peekaboo click --wait-for` already polls via `waitForElement`.

This change parses `wait_for` and, for `on` / `query` clicks, polls with the same `waitForElement` helper the CLI uses. Coordinate clicks still skip wait, matching CLI `--at`. Negative values are rejected.

## Evidence

Patched `ClickTool.execute` with `on` plus `wait_for: 8000`:

```console
unfixed: waitForElementCalls=[]
patched: waitForElementCalls=[elementId timeout=8.0 snapshot=<id>]
```

When the element is still missing after the wait, execute refuses with `targetUnavailable` instead of clicking.

`wait_for: -1` is rejected. Coordinate clicks still do not poll.

## Real behavior proof

- **Behavior or issue addressed:** Documented MCP `wait_for` was ignored, so click did not wait for the element to become actionable.
- **Real environment tested:** macOS arm64, patched PeekabooCore at `/tmp/peekaboo-f051`.

  ```bash
  # ClickTool.execute { on, wait_for: 8000 } against a snapshot
  ```

- **Evidence after fix:** terminal output from the patched tool:

  ```console
  unfixed: waitForElementCalls=[]
  patched: waitForElement timeout=8.0s for element id
  missing after wait: targetUnavailable, no click
  ```

- **Observed result after fix:** Element and query clicks poll for `wait_for` milliseconds (default 5000). Coordinates skip wait. Negative `wait_for` is rejected.
- **What was not tested:** A live UI control that appears after ~1s under a real MCP host (the production `waitForElement` path is the CLI helper).


steipete:
Thanks for catching the ignored wait_for argument. I reviewed the complete wait-and-dispatch path and am closing this implementation because it does not preserve the selected target contract.

The shared wait helper returns not-found without inspecting anything at timeout zero. For positive waits it can refresh against the pointer-selected application, but ClickTool subsequently resolves and dispatches from its original snapshot. A newly found element therefore does not prove that the dispatched element belongs to the same process generation and exact window; opaque element IDs can also be reassigned between traversals.

A complete repair needs fresh exact-target observations and stable matching through the eventual dispatch, including a defined policy for snapshot-local element IDs. I am retaining that unresolved API decision in the maintenance report rather than shipping this polling wrapper. No contributor revision is requested.

PR #758, opened 2026-09-20
fix(cli): bound mutation lock waits and preserve reservation ownership

SebTardif:
Desktop mutation lock acquisition now waits asynchronously with a 15-second monotonic deadline instead of blocking the actor indefinitely. CLI and MCP callers await acquisition, reserve ownership before suspending, and release in-flight reservations on every exit. Duplicate concurrent IDs cannot overwrite a pending token; cancelled acquisition can retry with the same ID.

Because preparation can now suspend, MCP revalidates its frozen target after acquiring the reservation and checks cancellation before dispatch. Refused or cancelled requests cancel their reservation; failed cleanup records recovery debt on the shared snapshot gate. Existing valid timing values remain supported, including a UInt64.max wait limit without deadline overflow. Embedded-coordinator guidance and the changelog are updated.

Validation uses synthetic targets and disposable real-flock directories. The original async proposal failed 22 new target/cancellation assertions. The final candidate passes 41 focused CLI tests, 20 MCP authority/ledger tests and four lock-deadline tests, including an already-cancelled free lock. Independent Codex review with complete lock and context source is clean through P2. Exact-head macOS and CodeQL CI must pass before merge.

Thanks @SebTardif for identifying the lock wait and contributing the original implementation.


SebTardif:
@clawsweeper re-review

> Reserve pending mutation state before awaiting acquisition (P2)

Done. `beginDurableMutation` sets `durableMutationAcquisitionInFlight` before the flock wait so a second caller cannot both observe `pendingDesktopMutation == nil`. Overlapping wait test passed in 0.057s and cancel leaves 0 pending files.

steipete:
Merged as fe813b6e480cafb601f9b22ea2c77ddc88af631c.

Bound asynchronous desktop mutation lock acquisition, reserve IDs before suspension and revalidate MCP targets after preparation. Refused/cancelled reservations use shared recovery-debt cleanup. The original async candidate failed 22 target/cancellation assertions; the final source passes 41 focused CLI tests, 20 MCP authority/ledger tests and four lock tests, including real contention, retry, duplicate IDs and already-cancelled free locks. Independent P2 review with full source context and exact-head CI passed.

PR #766, opened 2026-09-21
feat(homebrew): select architecture-specific CLI assets

vincentkoc:
Homebrew now selects the smaller arm64/x86_64 CLI archives when a published release provides the complete uploaded pair with valid GitHub SHA-256 digests. The tap downloads and verifies each archive. Universal-only releases remain supported; partial pairs, duplicate assets, and invalid digests fail before dispatch.

The request ID, matching downstream run receipt, and terminal wait remain intact. The tap prerequisites are merged, including openclaw/homebrew-tap#63 for thin→universal→thin transitions. Its pinned parser accepts the actual selector output for the Darwin-only pair.

The supplemental Core suite exposed one stale fixture after #756: an expired-snapshot test used an invalid reference and therefore exercised validation instead of absence. Use a valid but absent reference, retaining every stale-diagnostic assertion. This candidate also includes the main-branch inventory-test correction from #767.

Validation: all 13 focused Node tests and 31 native Bridge target tests passed. The Node tests execute the actual workflow Bash with a secretless fake GitHub CLI, including exact dispatch arguments and failure propagation. A synthetic cross-repository contract test passed against tap commit 2fae7f2f3b9f11770fb9f94a926a86ea667f6391. Independent Codex review with the current downstream source is clean through P2. Exact-head macOS, supplemental and CodeQL CI must pass before merge. No publication workflow, release, tag or formula update was dispatched.

Thanks @vincentkoc for the implementation.


vincentkoc:
The receiver-side transition is fixed in https://github.com/openclaw/homebrew-tap/pull/63 (landed commit: https://github.com/openclaw/homebrew-tap/commit/2fae7f2f3b9f11770fb9f94a926a86ea667f6391).

An explicit universal archive now replaces the canonical Darwin pair, preserving the formula's version metadata, installation code, resources, and future reconciliation. Plain thin-pair reconciliation, explicit target templates, and four-platform updates retain their existing paths.

Proof covers thin → universal → thin for both explicit input forms, repeated updates, and no write after a failed download. The actual tap updater and Homebrew `formula.install` also passed that transition in one isolated prefix using cached published signed v4.4.0 input and retained signed slices: ARM thin, universal with native ARM and Rosetta Intel launches, then Intel thin. Runtime bytes, modes, signatures, architectures, and version/source metadata stayed intact. This addresses the P1 finding and all three before-merge/rank-up items in https://github.com/openclaw/Peekaboo/pull/766#issuecomment-5763453735.

The caller head remains `359191e8a17536373169ac649f256816b639fea0`; no release or current formula pin was changed.

A separate CI blocker remains: [the full Core suite](https://github.com/openclaw/Peekaboo/actions/runs/35621754731/job/106406252666) reports three unchanged baseline issues: two `AppToolLifecyclePinningTests` quit/hide expectations and one `PasteToolTransactionGateTests` inventory-message expectation. The complete Core tree is byte-identical to the previously qualified producer tree; that earlier green CI ran filtered suites, so it is not an equivalent full-suite pass. The remaining full-safe and Swift analysis jobs are still active. No retry or Swift repair has been applied.

steipete:
Merged as be0c5f6ec37d7d786a0e563d863295fc632eb3e7.

Select the complete verified Darwin CLI archive pair for Homebrew, with universal-only fallback and the exact downstream dispatch receipt. Thirteen focused Node tests, 31 native Bridge tests and a synthetic cross-repository parser contract check passed. The full Core gate exposed an invalid expired-snapshot fixture after #756; a valid absent reference preserves all stale-diagnostic assertions. The final head passed all four supplemental groups, normal macOS CI and CodeQL. Independent P2 review included the current merged tap contract and is clean.

PR #767, opened 2026-09-22
test(mcp): align inventory refusal expectations with authoritative lookup

steipete:
The full Core suite fails three assertions after authoritative application lookup was added: quit/hide now attempt a direct lookup after incomplete broad enumeration, and a direct result without selector proof reports inventory unavailable.

Update those two test expectations to match the current planner contract. All refusal metadata, missing-receipt, no-dispatch, and zero-clipboard-side-effect assertions remain. Runtime code is unchanged.

The failing baseline is the full Core job at https://github.com/openclaw/Peekaboo/actions/runs/35621754731/job/106406252666 (3 issues in 3,251 tests). Focused SwiftFormat and diff checks pass; independent Codex review is scoped-clean through P2. Hosted macOS CI and the complete supplemental suite will validate the exact candidate. Local heavy builds were avoided under the maintenance load restriction; the Linux Crabbox allocation failed because the configured 160 GB disk is smaller than its image minimum.


steipete:
The four supplemental groups (Core, libraries/apps, dependencies, and full-safe) and normal macOS build/test jobs passed on the exact head. Swift CodeQL lost its hosted runner during the build: GitHub reports “The hosted runner lost communication with the server”; no Swift job log was uploaded and analysis did not start. I retried that failed job once on the same commit. Merge remains gated on successful analysis.

steipete:
Merged as 82ef8c32276a8ad25acad5bad65746cc1317ba93. The exact source passed normal macOS CI and all four supplemental validation groups. Swift CodeQL passed on attempt 2 after the first hosted runner lost communication.

For clarity, the squash body described the test scope incorrectly: this patch adjusts two existing inventory-refusal expectations, not the tool catalog. It expects the authoritative TextEdit lookup to occur before app-mutation refusal and expects the canonical inventory-unavailable message for paste. The no-dispatch assertions remain intact.

PR #769, opened 2026-09-22
fix(scroll): reject overflowing amounts before dispatch

steipete:
Extreme scroll amounts could trap in abs(Int.min), smooth tick multiplication, or CLI result arithmetic. Validate the mathematical range through one shared ScrollRequest check, before CLI runtime/focus and before service lane preparation or input. Ordinary signed and zero amounts retain their behavior; valid limits are derived from Int rather than an arbitrary policy cap.

The synthetic service regression terminated the original test process with signal 5 on Int.min. The fixed candidate passes 24 focused scroll service tests and 12 command parsing tests, covering invalid extrema, no preparation/input on rejection, and accepted signed/zero boundary values without executing enormous scrolls. Focused lint/format checks and independent Codex review through P2 passed. Documentation and Unreleased notes are updated; stale migration comments and a redundant rethrow were removed in the same area.

Exact-head macOS and CodeQL CI must pass before merge.


steipete:
Merged as b210983013f18baf8b962fa7249aa2371cbc6c6d.

Reject unrepresentable scroll amounts at the shared service and CLI pre-runtime boundary. Signed and zero amounts remain valid, while Int.min and overflowing smooth counts cannot reach abs, multiplication, focus or input. The original synthetic service test trapped with signal 5; all 24 focused service and 12 command-parsing tests passed with the fix. Independent P2 review, focused lint/format and exact-head macOS/CodeQL CI passed.

